PR #8, opened 2015-09-27
Replace classSet and joinClasses with classnames

dlwalsh:
I've replaced all references to `React.addons.classSet` and `React/lib/joinClasses` with [classnames](https://github.com/JedWatson/classnames).

The `classSet` addon is deprecated ([see here](https://facebook.github.io/react/docs/class-name-manipulation.html)) with `classnames` the recommended replacement.

Additionally, since `joinClasses` duplicates functionality provided by `classnames`, I've removed it as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,5 +41,8 @@
     "unmockedModulePathPatterns": [
       "<rootDir>/node_modules/react"
     ]
+  },
+  "dependencies": {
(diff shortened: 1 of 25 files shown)


isaiah:
Thanks. And sorry didn't see this before.

PR #14, opened 2015-11-06
Create a Badge component

stevegood:
I've created a simple Badge component based on the implementation examples here: http://materializecss.com/badges.html

### DIFF
--- a/src/Badge.js
+++ b/src/Badge.js
@@ -1,0 +1,23 @@
+import React from 'react';
+
+class Badge extends React.Component {
+  render() {
+    let classes = ['badge'];
+    if (this.props.text) {
+      classes.push(this.props.text);
+      classes = classes.reverse();
+    }
+
+    return (
+      <span className={classes.join(' ')}>
+        {this.props.children}
+      </span>
+    );
+  }
+}
+
+Badge.propTypes = {
+  text: React.PropTypes.string
+};
+
+export default Badge;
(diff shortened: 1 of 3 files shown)


src/Badge.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+class Badge extends React.Component {
+  render() {
+    let classes = ['badge'];
+    if (this.props.text) {
+      classes.push(this.props.text);
```
isaiah: Please use `classnames` (https://github.com/JedWatson/classnames) library to compose class names.
I am not sure if the text could be other than `new`, in that case it should be a boolean prop.

stevegood: Yeah, sorry, I was running on very little sleep,  'new' should definitely be a boolean prop not a string.  I can close the PR , make the changes and submit a new one.

stevegood:
Closing to make additional changes.

src/Badge.js
isaiah: You don't have to close this PR, if you just push your change to your branch it will show here.

isaiah:
Hi @stevegood, if you don't have time to make the change, I'll merge this and fix it myself, it could be a quite useful component.

stevegood:
I have the change made and additionally I added to the docs.  I'm about to submit a new PR for those changes.

PR #18, opened 2015-11-16
Input improvements

ziad-saab:
* Made `Input` component work properly for `<textarea>` by adding the appropriate `materializecss` class.
* Made `Input` component work properly for `<select>` elements:
  * Distinguish between "browser-default" and material select
  * If browser default, add the appropriate class and put the `<label>` before the input
  * If not browser default, don't display a label and run `$.fn.material_select` on the input
* Only adding the `validate` class if the user doesn't request for it to be removed. This is to preserve backwards-compatibility

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -1,9 +1,24 @@
 import React from 'react';
+import ReactDOM from 'react-dom';
 import constants from './constants';
 import cx from 'classnames';
 import uuid from 'node-uuid';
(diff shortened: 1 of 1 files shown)


ziad-saab:
@isaiah can you please take a look and comment when you have some time?

isaiah:
:+1:  Thanks.

PR #19, opened 2015-11-17
Input labels adjustment

ziad-saab:
Adjusted the `<label>` element in `Input` …

For more info, see the [forms documentation of materializecss](http://materializecss.com/forms.html#!),
specifically in the "Prefilling Text Inputs" section.

Also fixed labels for radio and checkbox inputs, where the `input-field` class was messing with label placement. materializecss' website shows the checkbox/radio examples without that class.

@isaiah awaiting your comments.

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -4,6 +4,10 @@ import cx from 'classnames';
 import uuid from 'node-uuid';
 
 class Input extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {hasValue: false};
+  }
   componentDidMount() {
     this._setupEl();
   }
@@ -20,12 +24,13 @@ class Input extends React.Component {
         $inputEl.material_select();
       }
       else {
-        var $labelEl = $(this.refs.labelEl);
-        if ($inputEl.val()) {
-          $labelEl.addClass('active');
-        }
-        else {
-          $labelEl.removeClass('active');
+        var previousHasValue = this.state.hasValue;
+        var actualHasValue = !!$inputEl.val();
+
+        if (previousHasValue !== actualHasValue) {
+          this.setState({
+            hasValue: actualHasValue
+          });
         }
       }
     }
@@ -66,7 +71,10 @@ class Input extends React.Component {
         C = 'input';
         inputType = type || 'text';
     }
-    let htmlLabel = <label ref="labelEl" htmlFor={id}>{label}</label>;
+    let labelClasses = {
+      active: this.state.hasValue
+    };
+    let htmlLabel = <label className={cx(labelClasses)} htmlFor={id}>{label}</label>;
     return (
       <div className={cx(classes)}>
         {this.props.type === 'select' && this.props.browserDefault ? htmlLabel : null}
--- a/src/Input.js
+++ b/src/Input.js
@@ -35,7 +35,7 @@ class Input extends React.Component {
   render() {
     let classes = {
       col: true,
-      'input-field': true
+      'input-field': this.props.type !== 'checkbox' && this.props.type !== 'radio'
     };
     let {placeholder, id, type, label, ...props} = this.props;
     constants.SIZES.forEach(size => {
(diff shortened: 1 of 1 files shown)


src/Input.js
```diff
@@ -6,23 +6,36 @@ import uuid from 'node-uuid';
...
+      var $inputEl = $(ReactDOM.findDOMNode(this.refs.inputEl));
+
+      if (this.props.type === 'select' && !this.props.browserDefault) {
+        $inputEl.material_select('destroy');
+        $inputEl.material_select();
+      }
+      else {
+        var $labelEl = $(ReactDOM.findDOMNode(this.refs.labelEl));
+        if ($inputEl.val()) {
+          $labelEl.addClass('active');
```
isaiah: This is not the best way to work with react, you should / can not modify DOM outside of react, for dynamic class names, always use the `classnames` function.

```javascript
import cx from 'classnames';

constructor(props) {
  super(this);
  this.state = { active: this.props.defaultValue !== null };
  this.onChange = this.onChange.bind(this);
}
render() {
  let labelClasses = {
     active: this.state.active,
  };
  let htmlLabel = <label className={cx(labelClasses)} htmlFor={id}>{label}</label>;

  // Listen on the change event of the input
  <C onChange={this.onChange} {...props} />
}
onChange(e) { this.state.active = $inputEl.val() !== null; }
```

src/Input.js
```diff
@@ -6,23 +6,36 @@ import uuid from 'node-uuid';
...
-    if (this.props.type === 'select' && !this.props.browserDefault && typeof $ !== 'undefined') {
-      var $el = $(ReactDOM.findDOMNode(this.refs.inputEl));
-      $el.material_select('destroy');
-      $el.material_select();
+    this._setupEl();
+  }
+
+  _setupEl() {
+    if (typeof $ !== 'undefined') {
+      var $inputEl = $(ReactDOM.findDOMNode(this.refs.inputEl));
```
isaiah: `ReactDOM.findDOMNode` is not necessary, the ref is the DOM now. see https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#dom-node-refs

src/Input.js
ziad-saab: @isaiah I tried to use `classnames` for this one too like I did with the rest, but it's not straightforward.

An `input` can get its value from different props like `value` or `valueLink` or even `defaultValue`. The only clean way I found to figure out for sure if an `input` has a value is to do it programatically after the component mounts.

Any suggestions on how to tackle this one would be appreciated :)

src/Input.js
ziad-saab: @isaiah fixing this one + another useless `findDOMNode`. Thanks :)

src/Input.js
isaiah: Yes, that's what I suggested, bind the class name to a state and modify the state when the value / valueLink / defaultValue changes. It might seem to be less straight forward at the beginning, but that's required, or else react lose control of the updates. I am pretty sure it triggers `invariant` warning here.

ziad-saab: @isaiah of course, how could I have missed this one. Fixing :)

ziad-saab: @isaiah fixed, please check when you have time :)

ziad-saab:
@isaiah if you don't mind I'd like to rebase in 1 commit before you merge it

src/Input.js
```diff
@@ -1,28 +1,45 @@
...
   componentDidUpdate() {
-    if (this.props.type === 'select' && !this.props.browserDefault && typeof $ !== 'undefined') {
-      var $el = $(ReactDOM.findDOMNode(this.refs.inputEl));
-      $el.material_select('destroy');
-      $el.material_select();
+    this._setupEl();
+  }
+
+  _setupEl() {
+    if (typeof $ !== 'undefined') {
```
ziad-saab: @isaiah in this case it's not on change though, it's only when the component is mounted, or if it receives new props. otherwise the `active` state is already managed by the underlying `materialize.js`.

Moreover in `_setupEl` I am only calling `setState` if there is a change is the `hasValue` attribute. This can happen at most once for `componentDidMount`, and once per new props received.

Thanks for the quick feedback :)

PR #23, opened 2015-12-05
Make 'select' type Input trigger onChange and update state correctly.

jareth:
Adding onChange to React select components doesn't work with Materialize's `<ul>` based virtual styled select component. See: https://github.com/Dogfalo/materialize/issues/1160

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -14,6 +14,10 @@ class Input extends React.Component {
   componentDidMount() {
     if (this.props.type === 'select' && !this.props.browserDefault && typeof $ !== 'undefined') {
       $(this.refs.inputEl).material_select();
+      $(this.refs.inputEl).on('change', this._onChange);
+      if (!!this.props.onChange) {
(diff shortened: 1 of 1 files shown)


src/Input.js
```diff
@@ -14,6 +14,10 @@ class Input extends React.Component {
   componentDidMount() {
     if (this.props.type === 'select' && !this.props.browserDefault && typeof $ !== 'undefined') {
       $(this.refs.inputEl).material_select();
+      $(this.refs.inputEl).on('change', this._onChange);
+      if (!!this.props.onChange) {
```
isaiah: You should move this to the `this._onChange` so that the existing function is not override.

isaiah:
I've fixed the lint, please rebase.

jareth:
- props.onChange call moved into this._onChange
- rebased

I noticed some of the other form components onChange aren't working as I would expect during testing this, but I would deal with them as a separate issue.

isaiah:
Thanks.

PR #35, opened 2016-02-05
Allow text color to change on CardPanel and remove default color from CardPanel 

mattsmith14:
The primary change allows the user to override the text color in the CardPanel. Removing teal was because I found it confusing how the CardPanel became teal in the first place.

Thanks!

### DIFF
--- a/src/CardPanel.js
+++ b/src/CardPanel.js
@@ -5,12 +5,11 @@ class CardPanel extends React.Component {
   render() {
     let classes = {
       'card-panel': true,
-      'teal': true
     };
     let {className, children, ...props} = this.props;
     return (
       <div className={cx(classes, className)}>
-        <span className="white-text">{children}</span>
+        <span>{children}</span>
       </div>
     );
   }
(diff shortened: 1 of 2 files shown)


isaiah:
Thanks for the PR.

Could you also change the `Card` examples to show how to set text color?

src/CardPanel.js
```diff
@@ -5,12 +5,11 @@ class CardPanel extends React.Component {
   render() {
     let classes = {
       'card-panel': true,
```
alextrastero: unnecessary comma :)

alextrastero:
Also, since it doesn't require any lifecycle, why not make it stateless?

mattsmith14:
@isaiah I have updated the example to show setting text color. Take a look and let me know if that is what you were going for.

@alextrastero I have removed the dangling comma, thanks for pointing that out. I agree on making the component stateless, but I can create a separate pull request for that.

Thanks!

mattsmith14:
I changed my mind and just went ahead and converted CardPanel to a functional component :-).

isaiah:
Thanks.

PR #37, opened 2016-02-14
change select input to re-materialize selects on update

arao456:
I was noticing an issue where if I changed the options passed into an `Input type='select'`, the underlying select would change but the wrapper would still display the old items. This should fix that, also off'd the listened for onChange, but let me know if that shouldn't be done.

Thanks!

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -12,9 +12,21 @@ class Input extends React.Component {
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


arao456:
@isaiah just realized there was a bug in this PR, the function `isMaterialSelect` doesn't actually return, I have no idea how I missed this. Will be making a new PR to fix this, let me know if this one should be reverted as well.

Sorry!

PR #40, opened 2016-02-19
isMaterialSelect function should return

arao456:
Fixes issue introduced in PR #37 

Let me know if I should revert that PR instead of just merging this one in.


isaiah:
Thanks.

Issue #41, opened 2016-02-22
Navbar doesnt apply color classname to parent


zareef4:
```
<Navbar className="blue" right>
    <NavItem` href=''>Foo</NavItem>
    <NavItem href=''>Bar</NavItem>
</Navbar>
```

If done as in the above example, the NavItems are colored blue but the rest of the navbar is still the default pink / red

isaiah:
#42 should have fixed this.

PR #42, opened 2016-02-22
Add classname prop to render

alextrastero:
Fixes issue: https://github.com/react-materialize/react-materialize/issues/41


isaiah:
👏

Issue #39, opened 2016-02-18
CardPanel does not accept inline styles


CoveTechnologyAus:
There seems to be an issue when wanting to change the width of CardPanel with inline styles.
For example, if I have this: 
```
<Row id="login-page">
          <Col s={12}>
            <CardPanel className="center" style={{width:300px'}}  >
         </Col>
</Row>
```
the CardPanel still expands to the whole width of the column and inspecting the element style in Chrome I notice there is no "width" style on the CardPanel. The other components I have tried seem to accept inline styles.

Any ideas?

isaiah:
The extra properties are not used for the `CardPanel` component, we could add that if really needed.

PR #43, opened 2016-02-23
re #39 append extra properties for CardPanel

isaiah:
@alextrastero

### DIFF
--- a/src/CardPanel.js
+++ b/src/CardPanel.js
@@ -12,7 +12,7 @@ const CardPanel = ({
 
   return (
       <div className={cx(classes, className)} {...props}>
-        <span>{children}</span>
+        {children}
       </div>
   );
 }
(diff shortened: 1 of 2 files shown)


alextrastero:
LGTM :+1:

src/CardPanel.js
```diff
@@ -11,7 +11,7 @@ const CardPanel = ({
   };
 
   return (
-      <div className={cx(classes, className)}>
+      <div className={cx(classes, className)} {...props}>
         <span>{children}</span>
```
alextrastero: This will produce an error if `children` is/are block elements

isaiah:
Do you mean we should remove the span element?

isaiah:
Do you mean we should remove the span element?

alextrastero:
Have you tried just adding {children} without a wrapper? If that doesn't work I would change `span` to `div`

isaiah:
@alextrastero removed, I think it makes sense.

isaiah:
@alextrastero If everything is ok, can you merge?

PR #45, opened 2016-03-03
Fix: select not update when receiving new props

scarletsky:


scarletsky:
In fact, I prefer to decouple `Select` / `Datepicker` from `Input`.

Any Ideas ?

isaiah:
That's a very good initiative, the select component is a beast of its own.

scarletsky:
@isaiah I will try to decouple the `Select` and `Datepicker` later.

PR #46, opened 2016-03-03
Add Select Component

scarletsky:
I will add test case later. And I will remove the `select` code in `Input.js`.


alextrastero:
Looking good :+1:

scarletsky:
I've write test case, but it can not run because the component require `jquery` and `materialize` as dependence.

I try to modify `karma.conf.js` and add:

```js
 files: [
      'https://cdnjs.cloudflare.com/ajax/libs/jquery/2.2.0/jquery.min.js',
      'https://cdnjs.cloudflare.com/ajax/libs/materialize/0.97.5/js/materialize.js',
      'test/*Spec.js'
 ]
```

but the output is :
```
04 03 2016 00:20:22.391:WARN [karma]: No captured browser, open http://localhost:9876/
04 03 2016 00:20:22.402:INFO [karma]: Karma v0.13.21 server started at http://localhost:9876/
04 03 2016 00:20:22.417:INFO [launcher]: Starting browser PhantomJS
04 03 2016 00:20:25.866:INFO [PhantomJS 1.9.8 (Mac OS X 0.0.0)]: Connected on socket /#L1QwLR66FBDM22VFAAAA with id 13602204
PhantomJS 1.9.8 (Mac OS X 0.0.0) ERROR
  Script error.


^C
```

How can I run my test cases?

isaiah:
@scarletsky Ideally, just `npm test`, could you push the change, so I can try locally?

scarletsky:
I pushed my commits.

If I do not change the `karma.conf.js` file, when I run `npm test`, it will throw:

```shell
ReferenceError: Can't find variable: $
```

or

```shell
TypeError: 'undefined' is not a function (evaluating '$(this.refs.__SELECT__).material_select()')
```

scarletsky:
I also notice that devDependencies miss the following package. Without these packages, `npm test` can not run.

```shell
├── chai@3.5.0  extraneous
├── karma-jquery@0.1.0  extraneous
├── lolex@1.4.0  extraneous
├── sinon@1.17.3  extraneous
└── sinon-chai@2.8.0  extraneous
```

isaiah:
@scarletsky those packages are introduced by `karma-sinon-chai`, `karma-jquery` is not necessary.

I've created a patch to fix the errors. https://gist.github.com/isaiah/c18b20f9c3629ebd6dec

scarletsky:
@isaiah 
No any reply on [gist](https://gist.github.com/isaiah/c18b20f9c3629ebd6dec) ?

scarletsky:
`karma.conf.js` should add the following files which mean `jquery` and `materialize-css` should be devDependences.

```js
    files: [
      'node_modules/jquery/dist/jquery.js',
      'node_modules/materialize-css/dist/js/materialize.js',
      'test/*Spec.js'
    ]
```

As `materialize.js` use `Function.bind`, we should add `bind-polyfill` or upgrade phantomjs to 2.x to run the test case.

alextrastero:
Hey @scarletsky any updates on this?

scarletsky:
@alextrastero 
I have no idea. If I want to run the test cases,  I have to add `bind-polyfill` or upgrade phantomjs to 2.x. 
But I am not sure whether to do it or not. If not, just close this pr.

alextrastero:
I see, thanks for the update. @isaiah, can you confirm?

isaiah:
Upgrade phantomjs sounds good to me.

alextrastero:
Trying again after
https://github.com/react-materialize/react-materialize/pull/53

alextrastero:
Ok, to make it work I had to install jquery which wasn't in the package.json. But still after that it came up with more errors which need to be fixed. Please rebase with master and try again :)

test/SelectSpec.js
```diff
@@ -0,0 +1,38 @@
...
+    let instance = TestUtils.renderIntoDocument(
+      <Select
+          s={12}
+          m={8}
+          l={6}
+          defaultValue="2"
+          onChange={() => {}}>
+        <option value="1">Option 1</option>;
+        <option value="2">Option 2</option>;
+        <option value="3">Option 3</option>;
```
alextrastero: These 3 semicolons are a problem.

src/Select.js
```diff
@@ -0,0 +1,83 @@
...
+    }, () => this.props.onChange && this.props.onChange(e));
+  }
+
+  render() {
+    const { children, label, browserDefault } = this.props;
+    const classes = Object.assign(
+      {},
+      {
+        'input-field': true,
+        'col': true,
```
alextrastero: for this col to take effect I think you need a row class on parent

src/Select.js
```diff
@@ -0,0 +1,83 @@
...
+
+  _onChange(e) {
+    this.setState({
+      value: e.target.value
+    }, () => this.props.onChange && this.props.onChange(e));
+  }
+
+  render() {
+    const { children, label, browserDefault } = this.props;
+    const classes = Object.assign(
```
alextrastero: This is a nice way to create the size classes but it seems there's another way to do this without having to install yet another package, wdyt?

src/Select.js
alextrastero: Ah ok, I guess you could write <Row> <Select > </Row> but is that documented somehow?

test/SelectSpec.js
scarletsky: Oh, I didn't realize I add semicolon in jsx.

src/Select.js
scarletsky: Personally, I really like to use `Object.assign` to create a new object.
But if we do not like to install another package, I can use another way to archive this.

scarletsky:
The test environment miss jquery ?

alextrastero:
npm i --save-dev jquery fixed it for me

scarletsky:
@alextrastero 
https://github.com/react-materialize/react-materialize/pull/46/commits/6b7234ceec8b37e0ff796d1990ae855dcd2d1be3
Unless I do this, otherwise I can't run the test...

karma.conf.js
```diff
@@ -16,6 +16,8 @@ module.exports = function(config) {
 
     // list of files / patterns to load in the browser
     files: [
+      'node_modules/jquery/dist/jquery.js',
+      'node_modules/materialize-css/dist/js/materialize.js',
```
alextrastero: strange, I didn't need this to make the specs work

alextrastero: but as you can see All checks are green :)

scarletsky:
I decide to close this pr as I do not have enough time to finish it... I am so sorry.

PR #64, opened 2016-05-10
Collection item href support

linuxenko:


alextrastero:
merged here: https://github.com/react-materialize/react-materialize/pull/65

PR #66, opened 2016-05-10
Clean up collapsible and collapsible_item

alextrastero:
A small PR to clean up the collapsible and collapsableItem components 🐰

### DIFF
--- a/src/Collapsible.js
+++ b/src/Collapsible.js
@@ -12,14 +12,13 @@ class Collapsible extends React.Component {
   }
 
   render() {
-    let {
-      accordion, popout, className, children, ...props
-    } = this.props;
-    let classes = {
+    const { accordion, popout, className, children, ...props } = this.props;
+    const classes = {
       collapsible: true,
-      popout,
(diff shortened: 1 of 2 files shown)


src/Collapsible.js
```diff
@@ -61,12 +60,14 @@ Collapsible.propTypes = {
...
   accordion: React.PropTypes.bool,
+  className: React.PropTypes.string,
+  children: React.PropTypes.node,
   /**
    * Enable popout style
    */
   popout: React.PropTypes.bool,
   defaultActiveKey: React.PropTypes.number,
-  onSelect: React.PropTypes.func,
+  onSelect: React.PropTypes.func
```
isaiah: Why don't you like the trailing comma?
It prevents errors when the list is extended, and also avoid introducing changes on the diff :)

alextrastero: it's a question of style... some people like double quotes... no semi-colon... and other. I agree about the diff though. I guess it's not a big deal either way :)

Issue #74, opened 2016-06-06
JS components not initially rendered don't work


rulatir:
Javascript components like Dropdown don't work if they are not rendered initially but only subsequently upon an ancestor's state change.

Attached is a diff against boilerplate commit 2fe4b89847a57459804d8078e8ffd98158492042 that demonstrates the issue. You need to change the initial `ready` state value from `true` to `false` to see the symptom.

[demonstrate-delayed-render-bug.txt](https://github.com/react-materialize/react-materialize/files/300235/demonstrate-delayed-render-bug.txt)

This is probably the real issue behind #70 .

sminutoli:
Dropdown isn't working yet, I'm using 0.14.5 because 0.14.6 throws errors...

alextrastero:
hey @sminutoli, what kind of errors are you seeing?

sminutoli:
Hi @alextrastero I'm getting a `jquery.js:4311 Uncaught TypeError: Cannot read property 'createDocumentFragment' of undefined`. 
After some debugging I've found that `this._trigger` is a ReactElement instead a DOM Element in the `componentDidMount` hook. 
If I apply `$('.dropdown-button').dropdown()` instead of `$(this._trigger).dropdown()` it seems to work well (tough isn't the right solution).
I'm having this issue mounting the dropdown _after_ a re-render of a previous screen/route.

sminutoli:
@alextrastero do you have a clue about this one?

sminutoli:
I've made a PR with a proposal https://github.com/react-materialize/react-materialize/pull/110

alextrastero:
please reopen if this is not fixed :)

PR #75, opened 2016-06-06
Dropdown Component - Initialize when mounted

alextrastero:
Related to https://github.com/react-materialize/react-materialize/issues/74

Added a idx ref and initialised dropdown when mounted

This fixes only the dropdown, more to come.

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -12,7 +12,7 @@ class Dropdown extends Component {
 
     $(this._trigger).dropdown(options);
   }
- 
+
   componentWillUnmount() {
     $(this._trigger).off();
   }
@@ -53,12 +53,12 @@ Dropdown.propTypes = {
    * more info: http://materializecss.com/dropdown.html#options
    */
   options: PropTypes.shape({
-    induration: PropTypes.number,
-    outduration: PropTypes.number,
-    constrainwidth: PropTypes.bool,
+    inDuration: PropTypes.number,
+    outDuration: PropTypes.number,
+    constrain_width: PropTypes.bool,
     hover: PropTypes.bool,
     gutter: PropTypes.number,
-    beloworigin: PropTypes.bool,
+    belowOrigin: PropTypes.bool,
     alignment: PropTypes.oneOf(['left', 'right'])
   })
 };
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -1,34 +1,45 @@
-import React from 'react';
+import React, { Component, PropTypes } from 'react';
 import idgen from './idgen';
 
-class Dropdown extends React.Component {
+class Dropdown extends Component {
   constructor(props) {
     super(props);
     this.renderTrigger = this.renderTrigger.bind(this);
   }
 
-  render() {
-    let {trigger, children, ...props} = this.props;
-    let idx = 'dropdown_' + idgen();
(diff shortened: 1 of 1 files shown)


alextrastero:
```
<Dropdown options={ { hover: true } } trigger={ this.renderTrigger() }>
```

alextrastero:
WIP, need to check if it still works

wedneyyuri:
@alextrastero Have you thought about memory leak possibilities since we are not removing the event listeners?

https://facebook.github.io/react/docs/component-specs.html#unmounting-componentwillunmount

alextrastero:
Good catch @wedneyyuri, i've update the PR, ready for codereview

wedneyyuri:
@alextrastero What do you think about removing the `beloworigin` option from props/defaultProps? I think we can pass this value inside the `options` props. 

Everything seems good to me, a good work :)

alextrastero:
@wedneyyuri yeah I thought about it, but IMO it's a must to have it `beloworigin` as default. but we still want to keep it customizable

src/Dropdown.js
```diff
@@ -1,50 +1,66 @@
...
 
-class Dropdown extends React.Component {
+class Dropdown extends Component {
   constructor(props) {
     super(props);
     this.renderTrigger = this.renderTrigger.bind(this);
   }
 
+  componentDidMount() {
+    const options = this.props.options || {};
```
wedneyyuri: What do you think about it?

```
const options = $.extend({
  belowOrigin: true
}, this.props.options);
```

src/Dropdown.js
```diff
@@ -1,50 +1,66 @@
...
+   * Options hash for the dropdown
    */
-  overorigin: React.PropTypes.bool,
+  options: PropTypes.object
 };
 
 Dropdown.defaultProps = {
-  overorigin: false,
-}
+  beloworigin: true
```
wedneyyuri: We can add our default value for options here:
`options: { }`

wedneyyuri:
My point is make all the customisation inside the same `options` object to keep the original behaviour from materializeCSS.

Another possibility is using destructors (maybe a breaking change):

`<Dropdown inDuration={300} hover={true} />`

```
componentDidMount() {
  const options = this.props;
  ...
}
```

alextrastero:
Yeah, the problem with this is that trigger, className or other props would also be passed to the dropdown

alextrastero:
I mean, if what we want is to keep the functionality like materialize then we should remove the beloworigin is just pass the `options={ hover: true, beloworigin: true }` as a prop

wedneyyuri:
I agree

alextrastero:
@wedneyyuri check out the latest commit :)

wedneyyuri:
@alextrastero Seems good to me.

alextrastero:
The documentation should be update to reflect these changes
 - beloworigin is not a proptype anymore
 - new proptype to be used like so:
```JS
      <Dropdown
        trigger={<span>Trigger me</span>}
        options={
          {
            inDuration: 123,
            outDuration: 321,
            constrain_width: true,
            hover: true,
            gutter: 10,
            belowOrigin: true,
            alignment: 'right'
          }
        }
      >
```

isaiah:
:rocket:

Issue #70, opened 2016-06-02
Nothing works with rect-router


nazarposhta:
Do you know that nothing works when someone want to use react-router and tries to navigate on the site by 
`<Link to="/some-link"></Link>`
Plugins (Tabs) or something was not initialized 
We are making the single page application - actually. But it's not possible with that package.

nazarposhta:
I resolved that with 
``componentDidMount(){
        $('ul.tabs').tabs();
    },
``

rulatir:
Please fix this. The workaround suggested by the OP is not feasible in a nontrivial app. Also other components are failing too (Dropdown!)

alextrastero:
Hey @nazarposhta please reopen this issue if it remains, thanks!

PR #76, opened 2016-06-07
Tabs Component - Add initialise after mount

alextrastero:
related to https://github.com/react-materialize/react-materialize/issues/70

Tabs now are initialised after mount

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -1,43 +1,74 @@
 import React from 'react';
 import cx from 'classnames';
+import idgen from './idgen';
 
 import Row from './Row';
 import Col from './Col';
 
 class Tabs extends React.Component {
+  componentDidMount() {
+    if (typeof $ !== 'undefined') {
+      $(this.refs[this.tabsIdx]).tabs();
+    }
+  }
+
+  renderChildren() {
+    return (
+      React.Children.map(this.props.children, (child, idx) => {
+        const {
+          title,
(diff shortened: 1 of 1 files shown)


src/Tabs.js
```diff
@@ -1,43 +1,74 @@
 import React from 'react';
 import cx from 'classnames';
+import idgen from './idgen';
 
 import Row from './Row';
 import Col from './Col';
 
 class Tabs extends React.Component {
+  componentDidMount() {
+    if (typeof $ !== 'undefined') {
```
wedneyyuri: If jQuery is not defined, then it will fail silently.

alextrastero: Yup, i'm afraid this happens in many places for now...

alextrastero: Would need to Throw an exception

wedneyyuri: You don't need to explicitly throw an exception:

image

Just believe that jQuery will be available on the page :)

alextrastero:
WIP, still checking some stuff, but as @wedneyyuri mentioned, it's best to show an error if there is an error.

isaiah:
@alextrastero looks good to me, what's left here?

alextrastero:
I need to have another look in the w/e

Alex

reesmanp:
PR #197 might be relevant here.

alextrastero:

PR #78, opened 2016-06-13
decode Buffer to string before output, fix #77

### LINKED ISSUE #77: message.trim is not a function
Hi,

I am trying to run your `docs` by running these commands:

```
npm run build
npm run start
```

However I got error `message.trim is not a function` TypeError in `dev.js` while running `npm run start`. Is 
there a way to fix that?

Thanks!

isaiah:
Related changes:

https://nodejs.org/api/buffer.html
https://nodejs.org/api/string_decoder.html

### DIFF
--- a/docs/dev.js
+++ b/docs/dev.js
@@ -1,16 +1,17 @@
 // Shameless copied from react-bootstrap
 
 import 'colors';
-import {
-  exec
-}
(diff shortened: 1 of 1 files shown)


alextrastero:
👍

PR #81, opened 2016-06-17
Fix to renderBody

evgenosiptsov:
Render body need to conside props.expaned, setted in Colllapsible component this:

```
if (this.props.accordion) {
  props.expanded = child.props.eventKey === this.state.activeKey;
  props.onSelect = this.handleSelect;
}

return React.cloneElement(child, props);
```

### DIFF
--- a/src/CollapsibleItem.js
+++ b/src/CollapsibleItem.js
@@ -43,7 +43,7 @@ class CollapsibleItem extends React.Component {
   }
 
   renderBody() {
-    if (!this.state.expanded) return;
+    if (!this.state.expanded && !this.props.expanded) return;
 
     const style = {display: 'block'};
     return (


wedneyyuri:
@EGedeon Would you mind adding a test for this component in this pull request?

alextrastero:
@wedneyyuri not sure if that's fair to ask, his change is very small. Would be great if you could @EGedeon but don't feel forced :)

wedneyyuri:
You are right @alextrastero :)

isaiah:
I think a better approach is to synchronize this state on `componentWillMount`, @alextrastero ?

isaiah:
@EGedeon Thanks for the PR, can you make the change I suggested?

alextrastero:
Yes, agree

alextrastero:
fixed in #184

PR #82, opened 2016-06-17
Update CollectionItem.js

EGedeon:
className support for possible

```
class="collection-item avatar"
```

### DIFF
--- a/src/CollectionItem.js
+++ b/src/CollectionItem.js
@@ -7,21 +7,23 @@ class CollectionItem extends React.Component {
     this.isAnchor = this.isAnchor.bind(this);
   }
   render() {
+    let {active, className} = this.props;
+    
     let classes = {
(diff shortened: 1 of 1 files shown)


src/CollectionItem.js
```diff
@@ -7,21 +7,23 @@ class CollectionItem extends React.Component {
     this.isAnchor = this.isAnchor.bind(this);
   }
   render() {
+    let {active, className} = this.props;
```
alextrastero: tiny, could be const.

EGedeon: So, imposibble to set className staying? :)

isaiah: Can you setup the exlint rules to enforce this?

isaiah: will refactor all once we have this enforced. merging.

PR #88, opened 2016-07-03
adding ref and focus method to Input component

velopert:
I added ref and focus method to the Input component so that I can programatically focus the input.

By doing this, I can 

```javascript
<Input name="username" label="username" ref={ ref => { this.usernameInput } }/>
/* codes... */
this.usernameInput.focus();
```
Is there any better way to do this without modfying the Component? 

Thanks!


velopert:
I am closing this due to wrong indentation..

PR #89, opened 2016-07-03
adding ref and focus method to the Input component

velopert:
I added ref and focus method to the Input component so that I can programatically focus the input.

By doing this, I can 

``` javascript
<Input name="username" label="username" ref={ ref => { this.usernameInput } }/>
/* codes... */
this.usernameInput.focus();
```

Is there any better way to do this without modfying the Component file? 

Thanks!

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -9,6 +9,7 @@ class Input extends React.Component {
     this.state = {value: this.props.defaultValue};
     this._onChange = this._onChange.bind(this);
     this.isSelect = this.isSelect.bind(this);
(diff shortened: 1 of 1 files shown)


Rafael-Miceli:
Hi guys, I'm trying to achieve focus on the Input component and I'm looking for this commit in the latest release and seems it's not there.

alextrastero:
Hey @Rafael-Miceli, I just published `react-materialize@1.0.17` because you're right it somehow disappeared :) let us know if this works :)

PR #99, opened 2016-08-14
Add className functionality to progress bar

alextrastero:
Currently the progress bar doesn't support another color via classname, fixed 🤗

### DIFF
--- a/src/ProgressBar.js
+++ b/src/ProgressBar.js
@@ -1,28 +1,24 @@
 import React, { Component, PropTypes } from 'react';
 import cx from 'classnames';
 
-class ProgressBar extends Component {
-  render () {
-    const { progress, className } = this.props;
-
-    let style;
-    const classes = {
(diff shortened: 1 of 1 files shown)


src/ProgressBar.js
```diff
@@ -1,33 +1,37 @@
...
       style = {
-        width: this.props.progress + '%',
+        width: progress + '%'
       };
     }
 
     return (
-      <div className={cx('progress', this.props.className)}>
-        <div className={cx(classes)} style={style}></div>
+      <div className={cx('progress lighten-5', className)}>
```
yanickrochon: Why `lighten-5`?

alextrastero: in the progress bar, the progress is marker in X color and the remaining progress is lighten-5 of X

yanickrochon: ah! I see! nice touch.

Issue #87, opened 2016-07-03
Unknown Props Warning in React 15.2


velopert:
React 15.2 has been released. When I use react-materialize with react 15.2, I see some errors like: 

```
bundle.js:4005 Warning: Unknown prop `node` on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in div (created by Row)

bundle.js:4005 Warning: Unknown prop `s` on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in div (created by Col)
```

I hope this gets fixed soon :)
Thank you.

derekargueta:
+1, I'm getting the same thing

koon-kai:
+ 1

jibe0123:
I have the same things !

nazarposhta:
https://monosnap.com/file/Mrwah7UWrflpty7Vsy3xKizSEMwJZj

alextrastero:
@nazarposhta i'll have a look over the weekend

nickbarry:
I'm getting the same error.

zhengpd:
Same warning here. Use `<Input s={3} />` and got "Warning: Unknown prop `s` on <input> tag. ". I guess this is caused by passing `s` prop to `input` component. There is a document about unknown prop warning at https://gist.github.com/jimfb/d99e0678e9da715ccf6454961ef04d1b

yanickrochon:
This should be trivial to fix, no?

alextrastero:
Should get fixed with this
https://github.com/react-materialize/react-materialize/pull/101

yanickrochon:
`Warning: Unknown props `trigger`, `onRequestHide` on <div> tag.`

on 

```js
<Modal
  header="Foo Model"
  trigger={
    <Button>Click Me!</Button>
  }
>
  <p>Some dialog content...</p>
</Modal>
```

alextrastero:
this is fixed in 0.15.0

yanickrochon:
Actually, the `Modal` component is still complaining about the `trigger` prop.

8ctopotamus:
I am getting this error when using the <Tabs> component.
react 15.3.1
react-materialize 0.15.1

alextrastero:
hey @8ctopotamus, please get the latest version of react-materialize, this shouldn't be an issue anymore

8ctopotamus:
Thanks @alextrastero! I updated and it works just fine!

[...]

PR #103, opened 2016-08-16
Cleanup props before rendering HTML element proposal

yanickrochon:
This is a solution to solve issue #87. This actually solves the three PhantomJS errors (or warnings) from the current test cases. Adding more props to filter, and integrating it to the affected components should solve the problem and render cleaner HTML.

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import constants from './constants';
+import cleanProps from './clean-props';
 import cx from 'classnames';
 import Icon from './Icon';
 import idgen from './idgen';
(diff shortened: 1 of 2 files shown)


alextrastero:
Hey @yanickrochon, thanks for the contribution, I would rather not add any deleting and fix the problem from the route. Which is done here https://github.com/react-materialize/react-materialize/pull/106
Have a look if you can, cheers

yanickrochon:
I understand, but then you have to manually parse and maintain all the different possible props that can be set, no? Isn't this a bit restrictive, forcing to wrap everything manually instead of letting them pass through?

alextrastero:
True, but the reason for adding this warning is so that you don't pass unnecessary props down the line. The only component complaining was the Button, and now it's fixed, so I think this is solved ;-)

yanickrochon:
No effect since #109

PR #109, opened 2016-08-16
0.15.0 Fixes (Input ...props and button onClick)

alextrastero:
By trying to fix the react props warning, I removed the ...props spread, this broke other functionality.
As mentioned here: https://facebook.github.io/react/docs/transferring-props.html#transferring-with-...-in-jsx, there's a better way to handle this.

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -36,7 +36,7 @@ class Button extends Component {
       return this.renderFab(cx(classes, className));
     } else {
       return (
-        <C className={cx(classes, className)}>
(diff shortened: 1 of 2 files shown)


yanickrochon:
This looks good to me! I tested it, and everything works fine so far, just as it did for 0.14.8. Unless others see more issues to resolve, this should be published ASAP.

yanickrochon:
Well, there's only this `Warning: Unknown props `trigger`, `onRequestHide` on <div> tag. Remove these props from the element.` but it's quite trivial and can be resolved later.

PR #110, opened 2016-08-16
issue#74: Changes the reference to provide a valid DOM element

sminutoli:
Uses the this.idx prop as a jQuery selector to find the DOM element on componentDidMount after the init call

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -35,7 +35,7 @@ class Dropdown extends Component {
     const { trigger } = this.props;
 
     return React.cloneElement( trigger, {
-      ref: (t) => this._trigger = t,
+      ref: (t) => this._trigger = `[data-activates=${this.idx}]`,
       className: 'dropdown-button',
       'data-activates': this.idx
     });


smeijer:
Would be awesome if this one becomes merged. We really need this.

alextrastero:
@sminutoli Would it be possible to fix the conflicts and write a some specs?

jhillhouse92:
What's the status with this?

sminutoli:
Sorry by the delay, we moved to `material-ui` due the lack of support here.
@alextrastero I can fix the merge and write some spec if someone reviews it

yanickrochon:
I am seriously considering trashing this and also move to Material-UI... so many problems with MaterializeCSS and React-materialize, it's insane! And not because a lack of contribution, but because contributions are ignored, instead of being discussed and acted upon, and the few commits are usually done in parallel to contributors, by project members. Same thing happens with MaterializeCSS. It's a shame.

alextrastero:
Pull requests are more than welcome @yanickrochon :)

PR #118, opened 2016-08-22
Render modal with context Fixes #117

### LINKED ISSUE #117: Pass react context to Modal
The actual implementation of the Modal doesn't work with react-router. Since it's a different tree, the context is not passed to the Modal. 

[React Bootstrap approach.](https://github.com/react-bootstrap/react-bootstrap/blob/3b33652f7fe424f129af6a178411cc9fe67ed988/src/ModalTrigger.js#L65-78)

The core team also suggests to use 
`ReactDOM.unstable_renderSubtreeIntoContainer(this, <Foo />, container)`
[As you can see here](https://github.com/facebook/react/issues/4706)

jhernandezme:


alextrastero:
👍 thanks

PR #123, opened 2016-08-28
Adding configurable on/off labels

ernestomancebo:
This is an enhancement to configure the fixed "on/off" labels for the switch button, in order to place a label in another language for non-english users and also if is required to set another message than "on/off" (i.e. enable/disable, active/inactive) and letting the "on/off" as default if it's not provided.

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -124,14 +124,14 @@ class Input extends Component {
       return (
         <div className='switch'>
           <label>
-            {offLabel || "Off"}
+            {offLabel || 'Off'}
             <input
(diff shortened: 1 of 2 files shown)


alextrastero:
Superb 👍 , just to be picky how about adding propType validation
```
  offLabel: PropTypes.string,
  onLabel: PropTypes.string,
```

ernestomancebo:
Done with that @alextrastero, also fixed the failing test case that I added. I'll add some more test cases in the meantime.

PR #144, opened 2016-09-06
Fix Pagination initial state and add specs

alextrastero:


src/Pagination.js
```diff
@@ -6,23 +6,30 @@ import PaginationButton from './PaginationButton';
...
+      activePage: activePage > 0 && activePage <= items
+        ? activePage : 1
     };
 
     this.renderButtons = this.renderButtons.bind(this);
     this._onClick = this._onClick.bind(this);
   }
 
+  componentWillReceiveProps ({ activePage }) {
+    const { items } = this.props;
```
isaiah: I think the `items` props might be updated as well.

alextrastero: yeah, good point

alextrastero: For now, I think we can merge this as is. I can't think of a scenario where the items would update externally.

alextrastero: cc @isaiah

isaiah:
rebase please.

isaiah:
@alextrastero ping?

alextrastero:
Yeah, need to revisit this

Issue #152, opened 2016-09-26
Support className passthrough


yanickrochon:
User code should be able to set extra `className` values to just about any component.

* [x] `Button`
* [ ] `Breadcrumb` should set to the nav element
* [x] `Card`
* [x] `CardPanel`
* [ ] `CardTitle` should set to the top level div element
* [ ] `Chip` should set to the top level div element
* [x] `Col`
* [x] `Collapsible`
* [ ] `CollapsibleItem` should set to the li element
* [ ] `Collection` should set to the list (i.e. ul / div) element
* [x] `CollectionItem`
* [ ] `Dropdown` should set on the span (top level) element
* [x] `Footer`
* [ ] `Icon` should set on the i element
* [ ] `Input` should set to the input / textarea / select element
* [ ] `MediaBox` should set on the img element
* [x] `MenuItem`
* [ ] `Modal` should set to the div (top level) element
* [ ] `NavItem` should set to the li element
* [x] `Navbar`
* [x] `Pagination`
* [x] `PaginationButton`
* [ ] `Preloader` should set to the div (top level) element
* [x] `ProgressBar`
* [x] `Row`
* [ ] `SearchForm` should set to the form (top level) element
* [ ] `SideNav` should set to the nav (top level) element
* [x] `Slide`
* [ ] `Spinner` should set to the div (top level) element
* [ ] `Tab` should set to the top level element *(not implemented component)*
* [x] `Table`
* [x] `Tabs`
* [ ] `Tag` should pass to the `Chip` component
* [ ] `Toast` should set to the button element

*`Overlay` and `OverlayTrigger` are special cases... no?

alextrastero:
Yes indeed @yanickrochon, I'm currently working on basic specs for each component, I'm also fixing the 0.15 warnings and I will also pass className props to the above mentioned, thanks for this!

PR #161, opened 2016-10-29
add classname to <Dropdown />

uryyyyyyy:
Hi!

I want to use dropdown with custom className(change color to grey).
but now, its not supported yet.

refs: https://github.com/react-materialize/react-materialize/issues/152

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -1,5 +1,10 @@
 import React, { Component, PropTypes } from 'react';
 import idgen from './idgen';
+import cx from 'classnames';
+
+const classes = {
+  'dropdown-content': true
+};
 
 class Dropdown extends Component {
   constructor (props) {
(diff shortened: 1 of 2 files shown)


src/Dropdown.js
```diff
@@ -18,14 +18,14 @@ class Dropdown extends Component {
...
-    const { children, trigger, ...props } = this.props;
+    const { children, trigger, className, ...props } = this.props;
     this.idx = 'dropdown_' + idgen();
     delete props.trigger;
 
     return (
       <span>
         { this.renderTrigger() }
-        <ul {...props} className='dropdown-content' id={this.idx}>
+        <ul {...props} className={`dropdown-content ${className}`} id={this.idx}>
```
isaiah: you can use the `cx` helper. e.g.
https://github.com/react-materialize/react-materialize/blob/master/src/CardPanel.js#L14

uryyyyyyy: thx! It's done.

isaiah:
Thanks for the fix.

PR #178, opened 2016-11-25
Fixed bug with dynamic children in Collapsible

mihaibogdan10:
A React component's children can sometimes be `null`, e.g. when adding children dynamically based on a condition. Added a filter before rendering them in Collapsible.

Example use case that triggered the bug:
```js
let item2 = null;
if (condition) {
    item2 = <BaseItem eventKey="item2"  />;
}

<Collapsible accordion>
    <BaseItem eventKey="item1" />
    {item2}
</Collapsible>
```

Also, cleared the `key` and `ref` passed to `React.cloneElement`, because [according to the docs](https://facebook.github.io/react/docs/react-api.html#createelement), `key` and `ref` will be preserved anyway.

### DIFF
--- a/test/CollapsibleSpec.js
+++ b/test/CollapsibleSpec.js
@@ -31,6 +31,35 @@ describe('<Collapsible />', () => {
     assert(wrapper.find('[data-collapsible="accordion"]').length, 'with a data attribute');
   });
 
+  it('should work with null dynamic children', () => {
+    let wrapper = mount(
+      <Collapsible accordion>
+        {null}
+        <CollapsibleItem header='First'>
+          Lorem ipsum dolor sit amet.
+        </CollapsibleItem>
+      </Collapsible>
+    );
+
+    assert(wrapper.find('[data-collapsible="accordion"]').length);
+  });
+
+  it('should pass the key props to its children', () => {
+    let wrapper = mount(
+      <Collapsible accordion>
+        <CollapsibleItem header='First' key="testKey">
+          Lorem ipsum dolor sit amet.
+        </CollapsibleItem>
+        <CollapsibleItem header='First' key="testKey">
+          Lorem ipsum dolor sit amet.
+        </CollapsibleItem>
+      </Collapsible>
+    );
+
+    // Should only render one of the two children
+    assert.strictEqual(wrapper.find('li').length, 1);
+  });
+
   describe('<CollapsibleItem />', () => {
     it('renders', () => {
       let wrapper = mount(
--- a/test/CollapsibleSpec.js
+++ b/test/CollapsibleSpec.js
@@ -45,19 +45,17 @@ describe('<Collapsible />', () => {
   });
 
   it('should pass the key props to its children', () => {
-    let wrapper = mount(
+    let collapsibleChildren = mount(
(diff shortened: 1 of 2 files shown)


alextrastero:
thanks @mihaibogdan10 ! would it be possible to add a small test case? You could also test that children are receiving key and ref :)

mihaibogdan10:
Testing the key/ref thing wasn't trivial. I figured I could test that having two children with the same key results in only one of them rendering. I couldn't come up with a way to test the ref, though.

But give the docs, and seeing that the key is indeed passed on, I don't think that's really needed.

mihaibogdan10:
My personal preference would be to take out the second test, frankly, because of the warning. Unless you have any other ideas on how this could be tested?

alextrastero:
This looks good, what warning?

mihaibogdan10:
When running the tests:
```
ERROR: 'Warning: flattenChildren(...): Encountered two children with the same key, `.$testKey`. Child keys must be unique; when two children share a key, only the first child will be used.
    in ul (created by Collapsible)
    in Collapsible (created by Unknown)
    in Unknown'
```

test/CollapsibleSpec.js
```diff
@@ -31,6 +31,35 @@ describe('<Collapsible />', () => {
...
+        <CollapsibleItem header='First'>
+          Lorem ipsum dolor sit amet.
+        </CollapsibleItem>
+      </Collapsible>
+    );
+
+    assert.strictEqual(wrapper.find('li').length, 1);
+  });
+
+  it('should pass the key props to its children', () => {
```
alextrastero: Not sure if this description is very meaningful reading the assertion, could you test if the child receives the key separately from testing that it doesn't render duplicates, since that's what it seems you're testing

alextrastero:
Also, have a look at this: http://airbnb.io/enzyme/docs/api/ReactWrapper/key.html

mihaibogdan10:
@alextrastero I did look at http://airbnb.io/enzyme/docs/api/ReactWrapper/key.html, but it didn't work in this case, because the key is put on React components rather than actual DOM elements.
I tried selecting all the sub-elements of the wrapper, and none of them had `key` attached to them.

I get your point about the description of the test, but really the only visible change to the DOM that I managed to test for was that it didn't render two children with the same `key`.

mihaibogdan10:
As I was writing the above comment, I realized that `Collapsible`'s children don't HAVE to be `CollapsibleItem`s, so I replaced them with a `<li>` and I was able to use enzyme's `.key()`. The thing is that the returned value is `.$testKey`, so I guess I'll check with contains rather that string equality.

test/CollapsibleSpec.js
```diff
@@ -31,6 +31,33 @@ describe('<Collapsible />', () => {
...
+    let collapsibleChildren = mount(
+      <Collapsible accordion>
+        <li key='testKey'>
```
isaiah: `li` should really be a `CollapsibeItem`

mihaibogdan10: @isaiah: No, it shouldn't. Read above comments. It's not possible to get the key that way in Enzyme.

Issue #159, opened 2016-10-23
CollapsibleItem : allow node type in header prop


jhernandezme:
CollapsibleItem only allow `PropTypes.string` but one of my use cases adds an icon along the header 
 `<CollapsibleItem  header={(<span>Settings <i className="material-icons right">arrow_drop_down</i></span>)}`

which causes a warning, maybe using `oneOfType([PropTypes.string, PropTypes.node])`

alextrastero:
Hi @jhernandezme, you can use the icon prop of CollapsibleItem,
`<CollapsibleItem header='First' icon='filter_drama'>`.

Cheers

jhernandezme:
Hi @alextrastero thanks! this does works but I can not pass down the icon class like 'right' for right alignment, so should be there an iconClass prop?

PR #179, opened 2016-11-26
Collapsible Item: Allow icon class prop

jhernandezme:
As mentioned in #159 , this allows to pass a custom className to the icon being rendered in the header.

### DIFF
--- a/test/CollapsibleSpec.js
+++ b/test/CollapsibleSpec.js
@@ -53,7 +53,7 @@ describe('<Collapsible />', () => {
     it('accepts icon props', () => {
       let wrapper = mount(
         <Collapsible accordion>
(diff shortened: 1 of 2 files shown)


test/CollapsibleSpec.js
```diff
@@ -53,13 +53,12 @@ describe('<Collapsible />', () => {
     it('accepts icon props', () => {
       let wrapper = mount(
         <Collapsible accordion>
-          <CollapsibleItem header='First' icon='filter_drama'>
+          <CollapsibleItem header='First' icon='filter_drama' iconClassName="right">
```
alextrastero: Small but let's keep quotes consistent, in this case single

jhernandezme:
Done!

alextrastero:
Great! Thanks

PR #185, opened 2016-11-28
if the dropdown trigger has a className it will now be preserved.

pr1ntr:
Hi, I was using this dropdown and found that the trigger would get its class list stripped. I made a small fix to make sure it keeps it if it exists.

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -41,9 +41,11 @@ class Dropdown extends Component {
   renderTrigger () {
     const { trigger } = this.props;
 
+    const className = typeof trigger.className === 'string' ? `${trigger.className} dropdown-button` : 'dropdown-button';
+
     return React.cloneElement(trigger, {
       ref: (t) => (this._trigger = `[data-activates=${this.idx}]`),
-      className: 'dropdown-button',
+      className,
       'data-activates': this.idx
     });
   }


src/Dropdown.js
```diff
@@ -41,9 +41,11 @@ class Dropdown extends Component {
   renderTrigger () {
     const { trigger } = this.props;
 
+    const className = typeof trigger.className === 'string' ? `${trigger.className} dropdown-button` : 'dropdown-button';
+
     return React.cloneElement(trigger, {
       ref: (t) => (this._trigger = `[data-activates=${this.idx}]`),
-      className: 'dropdown-button',
+      className,
```
alextrastero: Since we include the `classnames` module you could do something like this:
```js
const className = cx({
  'dropdown-button': true
}, trigger.props.className);
...
  className: className
...
```

alextrastero: and then in `test/DropdownButtonSpec.js` please add a `it('passes classname down to trigger')`

pr1ntr: So what spawned all this is that I was passing `<Button />` as the trigger, however since this is all inside a form the button press would always trigger submit.  If I add `e.preventDefault()` to the button handler then it stops the dropdown from working (I could have tried changing the button type probably). So, I am passing regular dom elements which don't have `props` so we can't assume that trigger is a React Element, though it is a node. Would you suggest we set up conditions for both?

alextrastero:

Issue #166, opened 2016-11-05
Smooth Scrolling Props with Collapsible


reesmanp:
This is definitely an enhancement and I can code this if there is any interest in this.

If there is a lot of content within a CollapsibleItem Accordion, clicking on the next item will collapse the currently open item and expand the clicked one (which is what we want) but it leaves the viewpoint where the user clicked on the item which may be quite a bit ways down the page forcing the user to scroll up to read from the top.

I think that we should be able to send props in that does a smooth scroll to the top of the item when it is clicked.

yanickrochon:
My view on this is that, while it may seem useful, this is user land in regard to this package, and there are many more priorities to implement, first.

reesmanp:
I agree that this isn't a priority right now but I do think that this should be part of the framework. As a user of the package, I would assume that this item should handle this inherently.

So priority should be low but not necessarily thrown into user land for implementation.

isaiah:
@reesmanp sounds interesting, if you can provide a PR :)

reesmanp:
@isaiah I'll get right on it.

reesmanp:
This seems to be working correctly.

PR #191, opened 2016-12-03
Implement auto scrolling on expand

reesmanp:
Issue #166


src/CollapsibleItem.js
```diff
@@ -67,24 +82,27 @@ class CollapsibleItem extends Component {
...
   /**
    * The node type of the header
    * @default a
    */
-  node: PropTypes.node
+  node: PropTypes.node,
+  /**
+   * The scroll behavior for scrollIntoView
+   */
+  scroll: PropTypes.string
```
isaiah: Shouldn't we put this option to the `Collapsible` component? So it could apply to all `CollapsibleItem`s. Also you can use `PropTypes.oneOf(['auto', 'instant', 'smooth'])` as the type.

isaiah:
Please solve the conflict :)

src/CollapsibleItem.js
reesmanp: The oneOf would be perfect, thanks for sharing. I'll add that. We wouldn't want to put this in the collapsible component because in my testing the scroll would fire before the collapsibleItem re-render and it wouldn't work as expected.

PR #197, opened 2016-12-20
Tabs are now more in line with spec.

reesmanp:
Currently it seems that no content is hidden properly with tabs.

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -60,7 +60,8 @@ class Tabs extends Component {
         </Col>
         {
           children.map((child, idx) =>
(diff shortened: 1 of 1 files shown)


Mardaneus86:
I've found that the reference of the ul tag is different from the element that is used with the jQuery tabs function. Because of this the tabs are not being initialised properly and therefore not hidden either. Matching up the ref seems to fix the issue. I've made a commit in [my fork](https://github.com/Mardaneus86/react-materialize/commit/b766bfe64d1dd586fe34aa73fe6d851294bd68ae) to test.

isaiah:
:+1:

Issue #102, opened 2016-08-16
Implement Autocomplete


yanickrochon:

isaiah:
This is a nice feature actually, would you expect something like this?

`<Autocomplete name='country' data={this.getData} />`

alextrastero:
this has been implemented

jderose9:
There doesn't appear to be any way to set an initial value for the input, such as when returning to edit a record. Is that correct?

alextrastero:
Hey @jderose9, would you expect this result? or the dropdown not opened?
![screen shot 2017-06-07 at 08 55 12](https://user-images.githubusercontent.com/4152819/26865906-13602c90-4b5f-11e7-9093-603174aedac4.png)

jderose9:
I would expect it to be closed. The use case is basically editing a record. For instance, user creats a record, filling in the autocomplete input, and then saves it. When the user later returns to the form and the record is loaded, it doesn't seem that I can assign the existing value to the input.

alextrastero:
I would then recommend having a if condition for that, if there's a record use a normal Input and if not use an Autocomplete, from the `materialize.js` if there's any value inside this input the dropdown will be triggered, so there's no way for us to not trigger it.

I could pass a don't autocomplete prop to autocomplete but then you would need to initialise the autocomplete on focus... don't like this approach

jderose9:
If the user edits the existing value, the autocomplete should still work, so dynamically swapping for a regular input wont work.

You are saying that if the underlying input's value is set programmatically, the drop down is still expanded?

alextrastero:
Yes, I added a screenshot above of what happens when you add value programmatically

parthsuresh:
Has this been implemented? The documentation at [react-materialize forms](https://react-materialize.github.io/#/forms) doesn't show this option.

alextrastero:
Hey, it's implemented but missing in the docs, `<Autocomplete ... />`

stewmystre:
Having issues with the drop down list not showing up properly when minLength={0}, it is acting like minLength={1} any ideas?

matas30rubliu:
would love to have an option to set ID for my Autcomplete element

PR #199, opened 2016-12-21
Initial commit for Autocomplete component.

reesmanp:
Issue #102

### DIFF
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -80,7 +80,13 @@ class Autocomplete extends Component {
       <Row>
         <Col offset={offset} className={cx('input-field', className, classes)} {...props}>
           {icon && this.renderIcon(icon, iconClassName)}
-          <input type='text' id={_id} className='autocomplete' onChange={this._onChange} />
+          <input
+            className='autocomplete'
+            id={_id}
+            onChange={this._onChange}
+            type='text'
+            value={this.state.value}
+          />
           <label htmlFor={_id}>{title}</label>
           {this.renderDropdown(data, minLength)}
         </Col>
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -10,15 +10,11 @@ class Autocomplete extends Component {
     super(props);
 
     this.state = {
-      value: '',
-      minLength: this.props.minLength
+      value: ''
     };
-
-    delete this.props.minLength;
 
     this.renderIcon = this.renderIcon.bind(this);
     this.renderDropdown = this.renderDropdown.bind(this);
-    this.renderUl = this.renderUl.bind(this);
     this._onChange = this._onChange.bind(this);
   }
 
@@ -26,35 +22,32 @@ class Autocomplete extends Component {
     return <Icon className={iconClassName}>{icon}</Icon>;
   }
 
-  renderDropdown (data) {
-    const { value, minLength } = this.state;
+  renderDropdown (data, minLength) {
+    const { value } = this.state;
 
-    if (minLength && minLength > value.length) {
+    if (minLength && minLength > value.length || !value) {
       return null;
     }
 
-    return value && Object.keys(data).map((key, idx) => {
-      const index = key.toUpperCase().indexOf(value.toUpperCase());
-      if (index !== -1 && value.length < key.length) {
-        return (
-          <li key={key + '_' + idx} onClick={(evt) => this.setState({ value: key })}>
-            {data[key] ? <img src={data[key]} className='right circle' /> : null}
-            <span>
-              {index !== 0 ? key.substring(0, index) : ''}
-              <span className='highlight'>{value}</span>
-              {key.length !== index + value.length ? key.substring(index + value.length) : ''}
-            </span>
-          </li>
-        );
-      }
-    });
-  }
-
-  renderUl (data) {
-    const drop = this.renderDropdown(data);
-    return drop
-      ? <ul className='autocomplete-content dropdown-content'>{drop}</ul>
-      : null;
+    return (
+      <ul className='autocomplete-content dropdown-content'>
+        {Object.keys(data).map((key, idx) => {
+          const index = key.toUpperCase().indexOf(value.toUpperCase());
+          if (index !== -1 && value.length < key.length) {
+            return (
+              <li key={key + '_' + idx} onClick={(evt) => this.setState({ value: key })}>
+                {data[key] ? <img src={data[key]} className='right circle' /> : null}
+                <span>
+                  {index !== 0 ? key.substring(0, index) : ''}
+                  <span className='highlight'>{value}</span>
+                  {key.length !== index + value.length ? key.substring(index + value.length) : ''}
+                </span>
+              </li>
+            );
+          }
+        })}
+      </ul>
+    );
   }
 
   _onChange (evt) {
@@ -72,6 +65,7 @@ class Autocomplete extends Component {
       m,
       l,
       offset,
+      minLength,
       ...props
     } = this.props;
 
@@ -88,7 +82,7 @@ class Autocomplete extends Component {
           {icon && this.renderIcon(icon, iconClassName)}
           <input type='text' id={_id} className='autocomplete' onChange={this._onChange} />
           <label htmlFor={_id}>{title}</label>
-          {this.renderUl(data)}
+          {this.renderDropdown(data, minLength)}
         </Col>
       </Row>
     );
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -1,0 +1,116 @@
+import React, { Component, PropTypes } from 'react';
+import cx from 'classnames';
+import Icon from './Icon';
+import Row from './Row';
+import Col from './Col';
+
+class Autocomplete extends Component {
+  constructor (props) {
+    super(props);
+
+    this.state = {
+      value: ''
+    };
+
+    this.renderIcon = this.renderIcon.bind(this);
+    this.renderDropdown = this.renderDropdown.bind(this);
+    this._onChange = this._onChange.bind(this);
+  }
+
+  renderIcon (icon, iconClassName) {
+    return <Icon className={iconClassName}>{icon}</Icon>;
+  }
+
+  renderDropdown (data) {
+    const { value } = this.state;
+
+    let li = null;
+    if (value.length) {
+      li = Object.keys(data).map((key, idx) => {
+        const index = key.toUpperCase().indexOf(value.toUpperCase());
+        if (index !== -1 && value.length < key.length) {
+          return (
+            <li key={key + '_' + idx}>
+              {data[key] ? <img src={data[key]} className="right circle"/> : null}
+              <span>
+                {index !== 0 ? key.substring(0, index) : ''}
+                <span className='highlight'>{value}</span>
+                {key.length !== index + value.length ? key.substring(index + value.length) : ''}
+              </span>
+            </li>
+          );
+        }
+      })
+    }
+
+    return li;
+  }
+
+  _onChange (evt) {
+    this.setState({ value: evt.target.value });
+  }
+
+
+  render () {
+    const {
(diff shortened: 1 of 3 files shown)


reesmanp:
Need to add Spec for this.

Also, would someone mind helping me test this? The DOM looks identical to the Materialize DOM for this component but the dropdown doesn't seem to be rendering.

I have the data object as: 

`let data = {
  "Apple": null,
  "Microsoft": null,
  "Google": 'http://placehold.it/250x250'
}`

And the component being called:

`<Autocomplete title="Test Title" data={data} iconClassName="prefix" icon="textsms"/>`

src/Autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+  className: PropTypes.string,
+  /*
+   * The title of the autocomplete component.
+   */
+  title: PropTypes.string,
+  /*
+   * An object with the keys of the items to match in autocomplete
+   * The values are either null or a location to an image
+   */
+  data: PropTypes.object.required,
```
alextrastero: this should be `PropTypes.object.isRequired`

src/Autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+    const _ul = 'autocomplete-content dropdown-content';
+
+    return (
+      <Row>
+        <Col s={s} m={m} l={l} offset={offset} className={cx('input-field', className)} {...props}>
+          {icon && this.renderIcon(icon, iconClassName)}
+          <input type="text" id={_id} className={_autocomplete} onChange={this._onChange}/>
+          <label htmlFor={_id}>{title}</label>
+          <ul className={_ul}>{this.renderDropdown(data)}</ul>
+          {children}
```
alextrastero: why do we need these `children` ?

src/Autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+      ...props
+    } = this.props;
+
+    const _id = 'autocomplete-input';
+    const _autocomplete = 'autocomplete';
+    const _ul = 'autocomplete-content dropdown-content';
+
+    return (
+      <Row>
+        <Col s={s} m={m} l={l} offset={offset} className={cx('input-field', className)} {...props}>
```
alextrastero: I would add these responsive sizes like we do for example in `input.js`, I would also remove them from the default props

src/Autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+          {icon && this.renderIcon(icon, iconClassName)}
+          <input type="text" id={_id} className={_autocomplete} onChange={this._onChange}/>
```
alextrastero: this is clean but no real need for a const `_autocomplete`

src/Autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+      })
+    }
+
+    return li;
+  }
+
+  _onChange (evt) {
+    this.setState({ value: evt.target.value });
+  }
+
```
alextrastero: 🎨  extra empty line

src/Autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+  renderIcon (icon, iconClassName) {
+    return <Icon className={iconClassName}>{icon}</Icon>;
+  }
+
+  renderDropdown (data) {
+    const { value } = this.state;
+
+    let li = null;
```
alextrastero: this could also be written like so:
```JS
    return value && Object.keys(data).map((key, idx) => {
```
a bit cleaner

src/Autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+          <label htmlFor={_id}>{title}</label>
+          <ul className={_ul}>{this.renderDropdown(data)}</ul>
+          {children}
+        </Col>
+      </Row>
+    );
+  }
+}
+
+Autocomplete.propTypes = {
```
alextrastero: how about adding another prop `maxLength` which would determine the input max length before triggering the dropdown? in our example it's pretty basic because it's only 3 elements but if this dropdown was for countries for example and you just typed `a`... 🤔

alextrastero: I know this is not in materialize.css but I think it's a plus

src/Autocomplete.js
```diff
@@ -0,0 +1,116 @@
...
+  offset: PropTypes.string
+};
+
+Autocomplete.defaultProps = {
+  s: 12,
+  m: 12,
+  l: 12
+};
+
+export default Autocomplete;
```
alextrastero: just as curiosity what editor do you use?

src/index.js
```diff
@@ -38,6 +39,7 @@ import Tag from './Tag';
 import Toast from './Toast';
 
 export {
+  Autocomplete,
```
alextrastero: 👍

src/Autocomplete.js
reesmanp: Yeah, I debated about this for a while and chose to go with this route. I can change it to the other way in my update.

src/Autocomplete.js
reesmanp: For the sake of generality. I can't really think of a good use case but I put it in.

src/Autocomplete.js
reesmanp: That's a terrific idea.

src/Autocomplete.js
reesmanp: I use WebStorm as my IDE. Why?

alextrastero: This will suppress Github's warnings about no new line at end of file http://stackoverflow.com/questions/16761227/how-to-make-intellij-idea-insert-a-new-line-at-every-end-of-file

src/Autocomplete.js
alextrastero: I see your point, but think of autocomplete as just an input with some magic functionality, I would remove this :)

src/Autocomplete.js
```diff
@@ -63,29 +62,32 @@ class Autocomplete extends Component {
...
 
     return (
       <Row>
-        <Col s={s} m={m} l={l} offset={offset} className={cx('input-field', className)} {...props}>
+        <Col offset={offset} className={cx(cx('input-field', className), classes)} {...props}>
           {icon && this.renderIcon(icon, iconClassName)}
-          <input type="text" id={_id} className={_autocomplete} onChange={this._onChange}/>
+          <input type="text" id={_id} className='autocomplete' onChange={this._onChange} />
           <label htmlFor={_id}>{title}</label>
           <ul className={_ul}>{this.renderDropdown(data)}</ul>
```
alextrastero: Sorry to be a pain, but I really don't think we need a variable for `_ul`, also could we move the entire UL inside {renderDropdown} ?

src/Autocomplete.js
```diff
@@ -63,29 +62,32 @@ class Autocomplete extends Component {
...
+    const sizes = { s, m, l };
+    let classes = {};
+    constants.SIZES.forEach(size => {
+      classes[size + sizes[size]] = sizes[size];
+    });
 
     return (
       <Row>
-        <Col s={s} m={m} l={l} offset={offset} className={cx('input-field', className)} {...props}>
+        <Col offset={offset} className={cx(cx('input-field', className), classes)} {...props}>
```
alextrastero: Couldn't this be:
`<Col offset={offset} className={cx('input-field', className, classes)} {...props}>`

src/Autocomplete.js
```diff
@@ -23,37 +24,35 @@ class Autocomplete extends Component {
...
-                {index !== 0 ? key.substring(0, index) : ''}
-                <span className='highlight'>{value}</span>
-                {key.length !== index + value.length ? key.substring(index + value.length) : ''}
-              </span>
-            </li>
-          );
-        }
-      })
+    if (minLength && minLength > value.length) {
+      return null
```
alextrastero: 💄 semi

src/Autocomplete.js
reesmanp: Didn't think that cx took more than two arguments. I guess I should look at the specs, huh?

src/Autocomplete.js
reesmanp: I'm doing the eslint and building the testing suite for this now, so this is already fixed in the next update.

src/Autocomplete.js
reesmanp: I was actually thinking about this last night. Looks weird not to.

reesmanp:
@alextrastero That is what my second comment is about. As far as I can tell, the DOM is exactly the same as materialize's but the dropdown is not rendering (even though it _is_ showing in the DOM via chromes developer tools) and I am not sure why.

src/Autocomplete.js
```diff
@@ -0,0 +1,124 @@
...
+import Row from './Row';
+import Col from './Col';
+
+class Autocomplete extends Component {
+  constructor (props) {
+    super(props);
+
+    this.state = {
+      value: '',
+      minLength: this.props.minLength
```
alextrastero: There's no real need to keep this `minLength` in the state

src/Autocomplete.js
```diff
@@ -0,0 +1,124 @@
...
+    };
+
+    delete this.props.minLength;
```
alextrastero: and no need to delete this prop here

src/Autocomplete.js
```diff
@@ -0,0 +1,124 @@
...
+    this.renderUl = this.renderUl.bind(this);
+    this._onChange = this._onChange.bind(this);
+  }
+
+  renderIcon (icon, iconClassName) {
+    return <Icon className={iconClassName}>{icon}</Icon>;
+  }
+
+  renderDropdown (data) {
+    const { value, minLength } = this.state;
```
alextrastero: again here I would grab `minLength` from the props

src/Autocomplete.js
```diff
@@ -0,0 +1,124 @@
...
+
+    if (minLength && minLength > value.length) {
```
alextrastero: how about `if (minLength && minLength > value.length || !value) {` then below you could return the entire big one

src/Autocomplete.js
```diff
@@ -0,0 +1,124 @@
...
+      return null;
+    }
+
+    return value && Object.keys(data).map((key, idx) => {
```
alextrastero:
```js
    // ...
    return (
      <ul className='autocomplete-content dropdown-content'>
        {
          Object.keys(data).map((key, idx) => {
    // ...
```

src/Autocomplete.js
```diff
@@ -0,0 +1,124 @@
...
+              <span className='highlight'>{value}</span>
+              {key.length !== index + value.length ? key.substring(index + value.length) : ''}
+            </span>
+          </li>
+        );
+      }
+    });
+  }
+
+  renderUl (data) {
```
alextrastero: no need for this function as stated above

src/Autocomplete.js
```diff
@@ -0,0 +1,124 @@
...
+      data,
+      icon,
+      iconClassName,
+      s,
+      m,
+      l,
+      offset,
+      ...props
+    } = this.props;
+
```
alextrastero: here you would need to delete `minLength` from props to not pass it to `<Col>`

alextrastero:
p.s. ping me on Gitter if you haven't got this working locally yet

src/Autocomplete.js
reesmanp: The props goes unused in the render function so unless I pass it as an argument to render the dropdown, ESLINT yells at me :(

alextrastero:
Hey @reesmanp, just fixed an issue where selecting a element from the dropdown wouldn't add it to the input value. Also added the corresponding spec, have a look when you can and we'll merge this! :) thanks!

reesmanp:
@alextrastero thanks so much for your help! It looks good to me.

BLRplex:
Please check if it's possible to merge?

alextrastero:
merged thanks @reesmanp

PR #202, opened 2017-01-03
Added SideNav

richmeij:
Hi guys!
I wanted to get this thing going, so this PR is mainly to see if I'm moving in the right direction with this. 

Whats in this PR?
- SideNav component
- SideNavItem component
- Added docs/example
- Added first basic test (Will add more tests when the implementation is agreed on)

What more?
- Fixed tests (removed only and fixed eslint error in Parallax)
- Upped eslint-plugin-promise version (got a missing peer dependency error)
- Set babel-eslint as parser for eslint (If that is an issue please let me know)

Questions:
- The trigger is implemented in the same fashion as for example DropDown. I'm not sure if this is the way to go for this component as well. Are the trigger and the sidenav always rendered together?

- I added `right` and `closeOnClick` as props for SideNav, but you can also supply them as options. Whats the preferred approach here? I thought it was convenient and clear if you can supply these things as props, but should I then also add the other options? And then skip options altogether?

- I added a userview property to SideNavItem. Should that be a separate component or something? Should that be part of this PR in the first place?

Thanks for any input!

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,6 +4,9 @@ extends:
   - standard
   - standard-react
 
+parser:
+  babel-eslint
+
 globals:
   Materialize: true
   React: true
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,9 +4,6 @@ extends:
   - standard
   - standard-react
 
-parser:
-  babel-eslint
-
 globals:
   Materialize: true
   React: true
--- a/src/Parallax.js
+++ b/src/Parallax.js
@@ -11,7 +11,7 @@ class Parallax extends Component {
     return (
       <div className={cx('parallax-container', className)}>
         <div className='parallax'>
-          <img src={imageSrc}/>
+          <img src={imageSrc} />
         </div>
       </div>
     );
--- a/src/SideNav.js
+++ b/src/SideNav.js
@@ -9,15 +9,12 @@ class SideNav extends Component {
 
   componentDidMount () {
     const { options = {} } = this.props;
-    $(this._trigger).sideNav({ options });
-  }
-
-  componentWillUnmount () {
-    $(this._trigger).sideNav('hide');
+    $(this._trigger).sideNav(options);
   }
 
   render () {
     const { children, ...props } = this.props;
+    delete props.id;
     delete props.trigger;
     delete props.options;
     return (
--- a/src/SideNav.js
+++ b/src/SideNav.js
@@ -1,21 +1,81 @@
-import React, { PropTypes } from 'react';
-import Icon from './Icon';
+import React, { Component, PropTypes } from 'react';
 import idgen from './idgen';
 
-const SideNav = ({ children, id = [`sidenav_${idgen()}`] }) =>
-  <nav>
-    <ul className='right hide-on-med-and-down'>{children}</ul>
-    <ul id={id} className='side-nav'>{children}</ul>
-    <a href='#' data-activates={id} className='button-collapse'>
-      <Icon>view_headline</Icon>
-    </a>
-  </nav>;
+class SideNav extends Component {
+  componentDidMount () {
+    const { options = {}, right, closeOnClick } = this.props;
+    $(this._trigger).sideNav({
+      ...options,
+      edge: right || options.edge,
+      closeOnClick: closeOnClick || options.closeOnClick
+    });
+  }
+
+  componentWillUnmount () {
+    $(this._trigger).sideNav('hide');
+  }
+
+  render () {
+    const { children, id = [`sidenav_${idgen()}`], ...props } = this.props;
+    this.id = id;
+    delete props.right;
+    delete props.closeOnClick;
+    delete props.trigger;
+    delete props.options;
+    return (
+      <span>
+        { this.renderTrigger() }
+        <ul id={id} className='side-nav' {...props}>
+          {children}
+        </ul>
+      </span>
+    );
+  }
+
+  renderTrigger () {
+    const { trigger } = this.props;
+    return React.cloneElement(trigger, {
+      ref: (t) => (this._trigger = `[data-activates=${this.id}]`),
+      'data-activates': this.id
+    });
+  }
+}
 
 SideNav.propTypes = {
-  children: PropTypes.node,
+  /**
+   * sidenav id. If none is passed, an id will be generated.
+   */
   id: PropTypes.string,
+  /**
+   * Component that is rendered to trigger the sidenav
+   */
+  trigger: PropTypes.node.isRequired,
+  /**
+   * Position the sidenav on the right edge of the screen
+   * @default false
+   */
   right: PropTypes.bool,
-  left: PropTypes.bool
+  /**
+   * Closes side-nav on anchor clicks
+   * @default false
+   */
+  closeOnClick: PropTypes.bool,
+  /**
+   * Options hash for the sidenav.
+   * More info: http://materializecss.com/side-nav.html#options
+   */
+  options: PropTypes.shape({
+    menuWidth: PropTypes.number,
+    edge: PropTypes.oneOf(['left', 'right']),
+    closeOnClick: PropTypes.bool,
+    draggable: PropTypes.bool
+  }),
+  children: PropTypes.node
+};
+
+SideNav.defaultProps = {
+  right: false,
+  closeOnClick: false
 };
 
 export default SideNav;
--- a/src/SideNav.js
+++ b/src/SideNav.js
@@ -2,13 +2,14 @@ import React, { Component, PropTypes } from 'react';
 import idgen from './idgen';
 
 class SideNav extends Component {
+  constructor (props) {
+    super(props);
+    this.id = props.id || `sidenav_${idgen()}`;
+  }
+
   componentDidMount () {
-    const { options = {}, right, closeOnClick } = this.props;
-    $(this._trigger).sideNav({
-      ...options,
-      edge: right || options.edge,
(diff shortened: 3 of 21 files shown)


reesmanp:
Since Navbar switches to SideNav on small screens, could you add a pass through prop for NavBar that sets closeOnClick for SideNav if passed?

src/SideNav.js
```diff
@@ -1,21 +1,81 @@
...
+    edge: PropTypes.oneOf(['left', 'right']),
+    closeOnClick: PropTypes.bool,
+    draggable: PropTypes.bool
+  }),
+  children: PropTypes.node
+};
+
+SideNav.defaultProps = {
+  right: false,
+  closeOnClick: false
```
reesmanp: I don't think these should be default props since if they are not sent, undefined is falsey.

richmeij: I added defaultProps so they would show up in the documentation. Is there another (better?) way to go about this?

test/ButtonSpec.js
```diff
@@ -5,7 +5,7 @@ import Button from '../src/Button';
 
 let wrapper = shallow(<Button waves='light'>Stuff</Button>);
 
-describe.only('Button', () => {
+describe('Button', () => {
```
alextrastero: Thanks for pointing this out but this should live in another PR, just like other unrelated issues to the SideNav, checking the PR now!

.eslintrc
```diff
@@ -4,6 +4,9 @@ extends:
   - standard
   - standard-react
 
+parser:
+  babel-eslint
```
alextrastero: not needed as per comment below

src/Parallax.js
```diff
@@ -11,7 +11,7 @@ class Parallax extends Component {
     return (
       <div className={cx('parallax-container', className)}>
         <div className='parallax'>
-          <img src={imageSrc}/>
+          <img src={imageSrc} />
```
alextrastero: this is not for this PR :)

src/SideNavItem.js
```diff
@@ -0,0 +1,74 @@
...
+  icon: PropTypes.string,
+  user: PropTypes.shape({
+    background: PropTypes.string,
+    image: PropTypes.string,
+    name: PropTypes.string,
+    email: PropTypes.string
+  })
+};
+
+export default SideNavItem;
```
alextrastero: don't forget to export this component also in `index.js`

src/SideNav.js
```diff
@@ -1,21 +1,81 @@
...
+    });
+  }
+
+  componentWillUnmount () {
+    $(this._trigger).sideNav('hide');
+  }
+
+  render () {
+    const { children, id = [`sidenav_${idgen()}`], ...props } = this.props;
+    this.id = id;
```
alextrastero: I would add this in the constructor, it feels more natural
```JS
  constructor (props) {
    super(props);
    this.id = props.id || `sidenav_${idgen()}`;
  }
```

alextrastero: or even better??
```JS
SideNav.defaultProps = {
  id: `sidenav_${idgen()}`
  ...
```

src/SideNav.js
```diff
@@ -1,21 +1,81 @@
...
+    const { options = {}, right, closeOnClick } = this.props;
+    $(this._trigger).sideNav({
+      ...options,
+      edge: right || options.edge,
+      closeOnClick: closeOnClick || options.closeOnClick
+    });
+  }
+
+  componentWillUnmount () {
+    $(this._trigger).sideNav('hide');
```
alextrastero: is this necessary? if it unmounts it's no longer in the DOM right?

src/SideNav.js
alextrastero: I guess if you just add this to the `propTypes` it's enough :)

src/SideNav.js
```diff
@@ -1,21 +1,81 @@
...
-    <a href='#' data-activates={id} className='button-collapse'>
-      <Icon>view_headline</Icon>
-    </a>
-  </nav>;
+class SideNav extends Component {
+  componentDidMount () {
+    const { options = {}, right, closeOnClick } = this.props;
+    $(this._trigger).sideNav({
+      ...options,
+      edge: right || options.edge,
```
alextrastero: `edge` is supposed to be a string of either 'left' or 'right' correct? but `right` is a boolean 🤔

src/SideNav.js
richmeij: I meant: the default values show up in the documentation only if you add defaultProps. If thats not needed then this can be removed.

test/ButtonSpec.js
richmeij: Allright. I see theres already a commit in master on this point.

src/SideNavItem.js
```diff
@@ -0,0 +1,74 @@
+import React, { Component, PropTypes } from 'react';
+import cx from 'classnames';
+
+class SideNavItem extends Component {
+
+  renderUserView = (user) => {
```
richmeij: done

src/SideNavItem.js
richmeij: done

src/SideNav.js
richmeij: You're right, but this is removed anyway :)

src/SideNav.js
richmeij: Personally, I think defining defaults for props like this is pretty straightforward. In fact, I didnt change this, it's already like this in master :) But, if I add it to defaultProps, the docs will show `sidenav_${idgen()}` as the default value for id. Is that accepted?

src/SideNav.js
richmeij: Unsure :) I will check it out. I added it because it's also implemented in for example Dropdown.

src/Parallax.js
richmeij: Allright. Shall I make a small PR for this? Cause it needs to be fixed for all tests to pass I guess :)

src/SideNav.js
alextrastero: True, wouldn't look good, your call to either add it in constructor or inline :)

.eslintrc
richmeij: removed it

src/Parallax.js
alextrastero: check out #205

src/SideNav.js
richmeij: I will add it to the constructor.

src/Parallax.js
richmeij: Awesome :)

alextrastero:
Looking good!

richmeij:
Im struggling a bit with the comment from @reesmanp about NavBar. I see that NavBar indeed falls back to SideNav on small screens. But it doenst use this (new) SideNav component. Should it do that? Or is that a change outside of this PR?

alextrastero:
Ah, forgot to look at that, will do now

alextrastero:
To be honest though would you use a `Navbar` and a `SideNav` together? @richmeij @reesmanp

richmeij:
I can imagine some usecases, but I'll leave it at your view on how you guys want React Materialize to be. :)

Other that that, I will go ahead and add some more unit tests.

richmeij:
Question: the properties of the userview (image, name, email) are clickable (they are links), but you cant really set their functionality when you use SideNavItem. Should that be addressed in this PR?

reesmanp:
[https://github.com/react-materialize/react-materialize/blob/89088739a98519eaf1849eba96a47f463898b056/src/Navbar.js#L54](url)

Had me confused. I thought it was rendering a side-nav component on smaller screens but it turns out it is not. I'll remove my requested changes. Thanks @richmeij @alextrastero

alextrastero:
@richmeij what if we create a userview component after this PR?

alextrastero:
would that make sense? is there much functionality involved?

richmeij:
Well, I wasnt sure. We can make a separate component ofcourse, it seems logical to do so. Its just that I didnt see any other use for it other than inside a SideNav (and I needed it to recreate the example from the materializecss site).

That said, it's not a lot of work to make it a separate component at all, and it would make for a cleaner separation IMHO.

alextrastero:
@richmeij nice! we could create the component but just for SideNav, not exposing it to other components :)

richmeij:
@alextrastero We could indeed! One question though: How would a user go about to add a userview to the sidenav? By using the (future) UserView component? Or by specifying <SideNavItem userview /> (like it is now?)

alextrastero:
Actually let's leave this whole clicking part out of this PR, for now, why don't we just send the values without clicks

richmeij:
Yup we can do that. In my opinion what we should do now is seperate the userview as a component, and dont do any clicks. Just display the data. (In this PR?)

I was just wondering how a userview should be added to a SideNav.

alextrastero:
your call, either in this PR or not :)

richmeij:
@reesmanp I find that after the changes to Collapsible, the docs are no longer working 😱  Any thoughts?

alextrastero:
@richmeij lemme have a look

richmeij:
@alextrastero Any update?

alextrastero:
Absolutely true @richmeij, I'm reverting that PR now

richmeij:
Thanks for confirming @alextrastero. I will check the changes on my side ASAP.

richmeij:
Ok, merged changes and added all tests. Let me know if there is anything that still needs work :) @alextrastero @reesmanp

alextrastero:
This is great, thanks for the contribution

Issue #157, opened 2016-10-11
Click-only FAB problem


nazarposhta:
I decided to make FAB with "Click-only FAB". Is that possible?
According to materialize doc.  In original version it's possible:

> Click-only FAB
> If you want to disable the hover behaviour, and instead toggle the FAB menu when the user clicks on the large button (works great on mobile!), just add the click-to-toggle class to the FAB.

reesmanp:
@nazarposhta This should work now. Please let us know so we can close this issue.

PR #203, opened 2017-01-05
Added orientation and click-to-toggle classes to FAB

reesmanp:
Issue #157


src/Button.js
```diff
@@ -112,11 +115,16 @@ Button.propTypes = {
    */
   onClick: PropTypes.func,
   tooltip: PropTypes.string,
-  waves: PropTypes.oneOf(constants.WAVES)
+  waves: PropTypes.oneOf(constants.WAVES),
+  /**
+   * Click-Only
```
alextrastero: Maybe we could have a bit more description here as this is what will show on the documentation

src/Button.js
```diff
@@ -112,11 +115,16 @@ Button.propTypes = {
...
+   */
+  clickOnly: PropTypes.bool
 };
 
 Button.defaultProps = {
-  node: 'button'
+  node: 'button',
+  clickOnly: false
```
alextrastero: I'm not sure we need a default prop for this guy, if he's not supplied the effect will be the same.

reesmanp: Fair enough. I was just being explicit with the default. I will remove this in my next update.

src/Button.js
reesmanp: Will do.

PR #213, opened 2017-01-28
Replaces Karma + Jasmine with Mocha + jsdom

alextrastero:
Hey there,
Our current test suites was pretty outdated so I switched to a faster stack with mocha and jsdom.

I would like some help though if possible, I'm struggling to add materialize.js to the global namespace. Currently i'm getting errors with `$(...).material_select() is not a function`

Help ?


alextrastero:
never mind, solved this :)

PR #215, opened 2017-01-29
Change jsdom setup in tests

wedneyyuri:
Add [jsdom-global](https://github.com/rstacruz/jsdom-global) to reduce the boilerplate to setup jsdom in our tests. Also, it will remove some unnecessary global variables.

Reference: https://github.com/react-materialize/react-materialize/pull/213


alextrastero:
Massive! thanks, I was struggling with this, much better now 👍

PR #226, opened 2017-04-02
imgSrc to imageSrc

tonywatson:
Incorrect component property name


alextrastero:
Generally `img` is understood to be the same as `image`, but thanks! :)

PR #232, opened 2017-04-30
New version

alextrastero:
Since master is now protected, a PR is required for every new package version. I also added a changelog to keep track of future releases


isaiah:
:+1:

PR #244, opened 2017-05-11
fix props.placement

davidsunsbk:


davidsunsbk:
Yes, the props should be placement instead of alignment. I've changed rest of the related "alignment" variable name to "placement" in <Slide.js> and fixed the prop validation. Now it should be working and the tests won't have any problem.

alextrastero:
Why not change L30 to `delete props.alignment;` ?

davidsunsbk:
Yeah, I was curious why it's there the first time I looked at the code. I thought maybe it's there for something else, so I avoided modifying that part. If you think deletion won't cause any potential problems, I can make another push.

alextrastero:
Actually what you did is good, I would just remove the `delete props.placement;` and you're done

davidsunsbk:
Sure, thanks!

alextrastero:
thanks again @davidsunsbk, if you use the project and want to help out please continue! there plenty of small issues to be fixed in the issues page :)

davidsunsbk:
Sure, overall it's a great project and it still has a long way to go compared to traditional Materialize framework. Hope it gets better and I'll see what I can do to help with my limited skills. :)

alextrastero:
we can all learn together! :)

PR #251, opened 2017-05-21
added check for this.props.value while setting active state on Input

brendancol:
I had a problem where on I was binding `Input` (type=text) to a value in a model:
```
<Input type="text" label="Value:" value={this.state.onsiteRenewableEnergyProductionTxt.toString()} />
```

on the initial load of the component, it would display with the `active` label class:
screen shot 2017-05-21 at 3 03 36 pm

but on subsequent renderings it would not have the `active` class.

screen shot 2017-05-21 at 3 02 23 pm

This is a small fix (maybe not the right place) which will check `this.props.value` and add the `active` class to label if it exists.

Let me know what yall think.  I would love to get this merged.  Thanks for the great project.

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -10,7 +10,7 @@ class Input extends Component {
     super(props);
 
     this.state = {
-      value: props.defaultValue
+      value: props.value || props.defaultValue
     };
 
     this._onChange = this._onChange.bind(this);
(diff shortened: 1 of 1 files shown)


isaiah:
@brendancol Thanks for the fix, I think a more proper way probably is:
https://github.com/brendancol/react-materialize/blob/a7730210b3f37359461462189917e9da68d40a8a/src/Input.js#L13

```js
this.state = props.value || props.defaultValue;
```

WDYT?

brendancol:
@isaiah yes that looks like a better place. I've updated accordingly

PR #252, opened 2017-05-21
Adds Divider, Section and Container components

alextrastero:


alextrastero:
Very true, thanks

isaiah:
@alextrastero Correct me if I'm wrong, isn't there a `export Component from '../file'` syntax that we can use, so we don't have to duplicate the name twice?

alextrastero:
Yup, that's valid

alextrastero:
@isaiah I like the way material-ui only exports one file by moving each component into a separate folder/index.js. That way you could import components individually WDYT?

isaiah:
That's hard to use IMO, people can still import individually by `import
react-materialize/lib/Col`, if they prefer. MaterialUI did that probably
because each component is relatively more complex than ours.

Alexander Odell <notifications@github.com> schrieb am Di., 23. Mai 2017,
09:02:

PR #255, opened 2017-05-26
Change paths to hashes

alextrastero:
So paths to work on static html sites (did work on webpack-dev-server) so moved to hashes


alextrastero:
this is required for the docs to work in static mode @isaiah

Issue #267, opened 2017-05-29
Unknown prop `options` on <nav> tag for NavBar


macsplan:
Hi, new to react so might be just me. Getting this issue when I create a NavBar:

```
bundle.js:1328 Warning: Unknown prop `options` on <nav> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in nav (created by Navbar)
    in Navbar (created by Header)
    in div (created by Header)
    in Header (created by App)
    in div (created by App)
    in App (created by RouterContext)
    in RouterContext (created by Router)
    in Router
```

```
import React from 'react';
import {Navbar, NavItem, Icon} from 'react-materialize'

class Header extends React.Component {

  constructor() {
    super();
  }

   render(){
     return (
      <div>
        <Navbar brand='logo' right>
          <NavItem href="/newPost">New Post</NavItem>
        </Navbar>
      </div>
    );
  }
};


export default Header;

```

Any ideas where I'm going wrong?

isaiah:
@macsplan looks like a bug, hold on, fix is on the way.

isaiah:
The fix is in master. https://github.com/react-materialize/react-materialize/pull/269

macsplan:
Thanks, can someone push a new release with the fix in? last release was in April

alextrastero:
@macsplan https://www.npmjs.com/package/react-materialize `a week ago` :)

PR #270, opened 2017-05-30
\#267 remove options from nav props

isaiah:


alextrastero:
This is a duplicate fix #269

PR #275, opened 2017-06-05
Fixes active input label overlapping number values

achntrl:
There was a bug when the input had value from the props and no state yet because it has not been changed by the user. The label would overlap the value inside the input

### Before : 
before
### After :
after

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -62,6 +62,7 @@ class Input extends Component {
     var types = {
       'checkbox': e.target.checked,
       'select-multiple': this.getMultipleValues(e.target),
+      'number': parseFloat(e.target.value),
       'default': e.target.value
     };
     const value = types[e.target.type] || types['default'];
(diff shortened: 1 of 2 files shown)


achntrl:
Hi @alextrastero 
I am not using `defaultValue` because I have a custom behaviour with the `onChange` callback (it looks up the value from the previous input and adds 1). This value is computed within a mobx store.

I'm using `react-materialize@^0.18.4` (from my yarn.lock)

src/Input.js
```diff
@@ -120,7 +121,8 @@ class Input extends Component {
         inputType = type || 'text';
     }
     let labelClasses = {
-      active: this.state.value || this.isSelect()
+      active: this.state.value || this.isSelect() || this.props.value ||
```
isaiah: Let's wrap this into a function i.e. `isActive()`, The problem here is not that `props.value` is not checked (https://github.com/react-materialize/react-materialize/blob/master/src/Input.js#L13), but rather in js `0` is `false`. Probably a better way to describe it is:
```js
isActive: () -> {
    this.state.value !== '' || this.isSelect()
}
```

achntrl:
I'm closing this one, I haven't been working on React for a while I'm not sure it's still relevant

Issue #143, opened 2016-09-06
Pagination is stateful, and the state doesn't always update when needed


rulatir:
Currently Pagination is stateful - it stores currentPage in the state. This state sometimes doesn't update when the corresponding property changes:

![image](https://cloud.githubusercontent.com/assets/677694/18264674/03baddfa-7412-11e6-8da3-776c777ec572.png)

Another case (note the nonsense state that means "page 2 of 1"; a component should take care to never hold nonsense state, therefore I consider this a bug):

![image](https://cloud.githubusercontent.com/assets/677694/18264882/0a74fb98-7413-11e6-99df-9341b6707765.png)

The workaround is to force re-rendering every time with `key={"paginator_"+this.paginatorSeq++}` but it's an ugly hack.

alextrastero:
Hey @rulatir, you mean that when you paginate the state doesn't update? For me it works fine, what version are you using? Also, I can see in your screenshot that you only have 1 item

rulatir:
I paginate through a "search store" that holds filter parameters sent to the listing's server endpoint; those parameters also include the page number for pagination. That same page number from the search store is also passed as activePage prop to the Pagination component.

I understand that `items` in `Pagination` parlance actually mean pages - is my understanding wrong? (If it were to *actually* mean "items", then `Pagination` would need to know the items-per-page value, and there's nothing about that in `Pagination` docs). It is perfectly valid to have just one page of results, and I believe it makes sense to display the pagination widget anyway in this case.

I edited the original report to emphasize the nonsensical state held by the component in the second case, and to clarify that these are two distinct screenshot of two different component states.

alextrastero:
I see, thanks for the explanation, would this work? #144

rulatir:
I think this is even worse. You look at `this.props` and silently apply a correction if they are nonsensical, hiding an error in application logic. You might have a different approach to this, but I would be inclined to at least issue a warning.

My code *never* passes "page 2 of 1" props to `Pagination` - please look at the props in the screenshots. Instead this state *emerges* within `Pagination` because `items` is taken directly from props but `activePage` is held in the state and isn't updated when the props change - there is no `componentWillReceiveProps` in `Pagination`.

The actual problem seems to be that `Pagination` does not support controlled operation and we can't compatibly switch to the `defaultValue` idiom. I will live with `key` bumping for now.

alextrastero:
I understand, the fact that I'm checking the props initially is so it renders nicely.
[...]

PR #278, opened 2017-06-15
Allow pagination to support controlled operation

viadrew:
In regards to issue #143


alextrastero:
created spec for that: https://github.com/react-materialize/react-materialize/pull/279

Issue #80, opened 2016-06-17
Collapsible's accordion and popout does not work


evgenosiptsov:
HI!

`<Collapsible popout={true} accordion={true} onSelect={this.props.onSelect}>
            {this.props.sections.map(function(section) {
                return (
                    <CollapsibleItem header={this.props.renderHeader(section)}>
                        {this.props.renderContent(section)}
                    </CollapsibleItem>)
            }.bind(this))}          
            </Collapsible>`   
- With popout=true active li does not have 'active' className
- With accordion=true selected items doest not explained

Check please collapsible component, there is enought of bugs.

EGedeon:
https://github.com/react-materialize/react-materialize/pull/81

yanickrochon:
This component is broken, this is why it is not working. For example, the materialize example shows that the following HTML markup

```html
<ul class="collapsible popout" data-collapsible="accordion">
  <li>
    <div class="collapsible-header"><i class="material-icons">filter_drama</i>First</div>
    <div class="collapsible-body"><p>Lorem ipsum dolor sit amet.</p></div>
  </li>
  <li>
    <div class="collapsible-header"><i class="material-icons">place</i>Second</div>
    <div class="collapsible-body"><p>Lorem ipsum dolor sit amet.</p></div>
  </li>
  <li>
    <div class="collapsible-header"><i class="material-icons">whatshot</i>Third</div>
    <div class="collapsible-body"><p>Lorem ipsum dolor sit amet.</p></div>
  </li>
</ul>
```

Should render as

![selection_101](https://cloud.githubusercontent.com/assets/78461/19970042/2954626c-a1b1-11e6-84fe-816d681f4db7.png)

with this DOM structure

![selection_104](https://cloud.githubusercontent.com/assets/78461/19970125/77a8299e-a1b1-11e6-913f-9da6698e5eec.png)

However, the JSX markup

```jsx
<Collapsible popout>
  <CollapsibleItem header='First' icon='filter_drama'>
    Lorem ipsum dolor sit amet.
  </CollapsibleItem>
  <CollapsibleItem header='Second' icon='place'>
    Lorem ipsum dolor sit amet.
  </CollapsibleItem>
  <CollapsibleItem header='Third' icon='whatshot'>
    Lorem ipsum dolor sit amet.
  </CollapsibleItem>
</Collapsible>
```

despite rendering the same visual, has this DOM structure

[...]

PR #286, opened 2017-06-23
Collapsible/CollapsibleItem Fixes

alextrastero:


alextrastero:
Fixes #80

src/CollapsibleItem.js
```diff
@@ -1,86 +1,31 @@
...
+  header,
+  children,
+  icon,
+  iconClassName,
+  node: Node,
+  onSelect,
+  ...props
+}) => (
+  <li className={cx(className, { active: expanded })} {...props}>
+    <Node
```
isaiah: here should be `node`.

src/CollapsibleItem.js
```diff
@@ -1,86 +1,31 @@
...
+      className={cx('collapsible-header', { active: expanded })}
+      onClick={() => onSelect(eventKey)}
```
isaiah: we still need the default `handleClick` function to wrap the custom `onSelect` callback, and handle the toggle of `expanded` state. Overall this cannot be a stateless component, the `expanded` state has to be hold.

src/CollapsibleItem.js
alextrastero: `node` is renamed to `Node` in the deconstruct

src/CollapsibleItem.js
alextrastero: this is handled by parent component `Collapsible` L56 no?

isaiah: Hmm, I'm not sure any more, can you try on the docs page and see if it work?

alextrastero:
Nah, it's not working well, checking further

alextrastero:
ping @isaiah works like a charm now :)

isaiah:
@alextrastero there is a failing spec.

alextrastero:
because it's not working like a charm... :(

alextrastero:
@mattBlackDesign would you mind testing the new changes?

mattBlackDesign:
Hey @alextrastero I noticed some issues with `defaultActiveKey`. 

When `Collapsible` takes the following form it works as expected:

```
<Collapsible accordion popout defaultActiveKey={0}>         
  <CollapsibleItem header='First'>A</CollapsibleItem>
  <CollapsibleItem header='Second'>B</CollapsibleItem>
</Collapsible>
```

However, when `Collapsible` has children inside of it that are components that render `CollapsibleItem`, `defaultActiveKey` does not expand the specific element. 

i.e. 

```
<Collapsible accordion defaultActiveKey={0}>
  <ChildComponentWithCollapsibleItem />
  <ChildComponentWithCollapsibleItem />
</Collapsible>

...

class ChildComponentWithCollapsibleItem extends Component {
  ...
  render () {
    return(
      <CollapsibleItem header='Element'>Element</CollapsibleItem>
    )
  }
}
```

alextrastero:
In this case `ChildComponentWithCollapsibleItem` should pass all props to children:
```js
class ChildComponentWithCollapsibleItem extends Component {
  ...
  render () {
    const { children, {...other } = this.props;
    return(
      <CollapsibleItem header='Element' {...other}>{children}</CollapsibleItem>
    )
  }
}
```

PR #298, opened 2017-07-14
Turn button into link if href is present

ruediste:
I am using Buttons as links. This can already be accompished by writing
`<Button floating large className='red' waves='light' icon='add' node="a" href="../account/_add"/>`
This patch alllows to omit the node property.

### DIFF
--- a/docs/src/pages/ButtonsPage.js
+++ b/docs/src/pages/ButtonsPage.js
@@ -10,6 +10,7 @@ import raisedButton from '../../../examples/RaisedButton';
 import fixedActionButton from '../../../examples/FixedActionButton';
 import floatingButton from '../../../examples/FloatingButton';
 import horizontalFab from '../../../examples/HorizontalFAB';
+import buttonAsLink from '../../../examples/ButtonAsLink';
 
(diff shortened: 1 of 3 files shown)


alextrastero:
Hey @ruediste, I'm not sure about this PR, I would rather not add more logic to the button if we can help it

ruediste:
Hello @alextrastero. I think rendering buttons as links is important to allow opening the target page in a new tab. To reduce logic in the Button, perhaps pulling out the FAB would be an idea? Could be moved to <Button.FAB ... />. Talking of, what do you thing about a <Button.Link ... />?

isaiah:
@ruediste I can see the usage of it, probably keep the `node` prop when specified?

ruediste:
Good point, just added. I also wrote a test

alextrastero:
I still think this is adding way to much logic to `<Button>` just to avoid adding 1 prop

ruediste:
Thanks for the quick review of the pull request. I'll probably include the logic in a custom wrapper. I added a sample to the documentation instead.

isaiah:
@ruediste :+1:

PR #305, opened 2017-07-19
Property Customizer

ruediste:
I repeatedly had the need to customize the Button behavior. Therefore I added a propertyCustomizer which can be used to globally change how the properties are computed. This allows simple stuff such as setting a default waves effect or color, as well as adding a default onClick callback for links.


alextrastero:
And the `href` part of the PR is only to avoid adding one prop to `<Button />`

ruediste:
Thanks for your suggestion, will go on in that direction.

alextrastero:
But keep up the PR's! we need people like you!! :)

PR #310, opened 2017-07-21
Feature tooltip

PokeGuys:
Should I delete the html option or just keep it?
Since data-tooltip only accept html string rather than a React element.

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -1,0 +1,48 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+class Tooltip extends Component {
+  componentDidMount () {
+    $(this._tooltipped).tooltip();
+  }
+
+  componentWillUnmount () {
+    $(this._tooltipped).tooltip('remove');
+  }
+
+  render () {
+    const {
+      children,
+      tooltip,
+      delay,
+      position,
+      html
+    } = this.props;
+
+    return React.cloneElement(children, {
+      'data-tooltip': tooltip,
+      'data-delay': delay,
+      'data-position': position,
+      'data-html': html,
+      'ref': (obj) => { this._tooltipped = obj; },
+      className: cx(children.props.className, 'tooltipped')
+    });
+  }
+}
+
+Tooltip.propTypes = {
+  children: PropTypes.element.isRequired,
+  tooltip: PropTypes.string.isRequired,
+  delay: PropTypes.number,
+  html: PropTypes.bool,
+  position: PropTypes.oneOf(['top', 'right', 'bottom', 'left'])
+};
+
+Tooltip.defaultProps = {
+  delay: 350,
+  position: 'bottom',
+  html: false
+};
+
+export default Tooltip;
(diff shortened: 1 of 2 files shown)


src/Tooltip.js
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+class Tooltip extends Component {
+  componentDidMount () {
+    $('.tooltipped').tooltip();
```
isaiah: Can you save the `ref` to not pollution the namespace?

pokeguys: You mean using 'ref': (obj) => { this._tooltipped = obj; } to specify the element?
It will be buggy when the child is created by other component.
The data-tooltip-id will be placed outside of element props.

test/TooltipSpec.js
```diff
@@ -0,0 +1,56 @@
...
+
+describe('Tooltip', () => {
+  it('should contain class tooltipped', () => {
+    const { wrapper } = setup({ tooltip: '123' });
+    assert.ok(wrapper.hasClass('tooltipped'));
+  });
+
+  it('should convert into child node', () => {
+    const { wrapper } = setup({ tooltip: '123' });
+    assert.ok(wrapper.find('button').length);
```
alextrastero: To specifically see if the wrapper is a button, we could assert like so:
`assert.equal(wrapper.type(), 'button');`

test/TooltipSpec.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  it('should contain child classes', () => {
+    const { wrapper } = setup({ tooltip: '123' });
+    assert.ok(wrapper.find('button').hasClass('btn'));
+  });
+
+  it('should contains correct props and transfer into corresponding data-*', () => {
```
alextrastero: `it('should contain...` or `it('contains correct...` :)

src/Tooltip.js
alextrastero: Hey @pokeguys could you please elaborate?

pokeguys: e.g. I am using a custom component as child which is called `<Button>`.
```
import { Button, Tooltip } from 'react-materialize;
<Tooltip><Button waves='light'>test</Button></Tooltip>
````
It will place `data-tooltip-id` into the component but not inside the props and not functioning.
It only happens when you are not using html element as a child.
But in `$('.tooltipped').tooltip()` way, it will execute the code and show it properly.
```
context: {}
**data-tooltip-id: "1db47f5e-5064-450b-35d8-9086538c32ab"**
props: {
  children: "Test"
  data-delay: 30
  data-html: false
  data-position: "top"
  data-tooltip: "Test"
  ...
}
```

alextrastero:
Also could you please add `Tooltip` to the `index.js` too?

src/Tooltip.js
alextrastero: Ah ok, I would then leave the className as you suggested

isaiah:
@pokeguys please re-export `Tooltip` through `src/index.js`, or else it's not added to the root namespace.

alextrastero:
ping @pokeguys

alextrastero:

PR #312, opened 2017-07-23
Fix radio button documentation

ruediste:
The source code of the documenation of the radio button was placed on the right of the sample. Replacing the containing div with a row (like all other samples) fixes the issue.


ruediste:
changing remote branch

PR #329, opened 2017-08-11
Fixes Tabs init with multiple instances + tabOptions

### LINKED ISSUE #325: Unable to Use Two <Tabs/> on the Same Page
To reproduce, simply 

```
<Tabs className='tab-demo z-depth-1'>
		<Tab title="Test 1">Test 1</Tab>
		<Tab title="Test 2" active>Test 2</Tab>
</Tabs>

<Tabs className='tab-demo z-depth-1'>
		<Tab title="Test 1">Test 1</Tab>
		<Tab title="Test 2" active>Test 2</Tab>
</Tabs>
```

It seems the <Tab/> element renders with an `id` to be `tab-0`, `tab-1`, `tab-2`, ... Therefore, the scope is the same for both elements.

alextrastero:
Fixes #325

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import idgen from './idgen';
 import cx from 'classnames';
 
 import Row from './Row';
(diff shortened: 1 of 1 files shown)


alextrastero:
Yes sure it is :)

PR #336, opened 2017-08-18
Feature/card horizontal

muddydixon:
## Summary
* Support Card horizontal

## How to Check
* Open `/#/cards` in doc.


alextrastero:
Thanks @muddydixon, could you also add a spec for this? if prop is provided XYZ dom should render 👍

muddydixon:
Write tests. plz check it.

PR #349, opened 2017-08-30
Fixing NavItem rendering

alextrastero:
restored NavItem original behaviour... I'm not sure it will work for all scenarios though...

### DIFF
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -1,26 +1,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const renderChildren = (children, href = '') => {
-  if (React.Children.count(children) === 1 && typeof children === 'string') {
-    return <a href={href}>{children}</a>;
-  } else {
-    return React.Children.map(children, (child) =>
-      React.cloneElement(child)
-    );
-  }
-};
-
 const NavItem = ({
   divider,
-  href,
   children,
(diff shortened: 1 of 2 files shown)


src/NavItem.js
```diff
@@ -1,26 +1,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const renderChildren = (children, href = '') => {
-  if (React.Children.count(children) === 1 && typeof children === 'string') {
-    return <a href={href}>{children}</a>;
-  } else {
-    return React.Children.map(children, (child) =>
```
isaiah: What I mean is we keep the `renderChildren` function, but remove this `map & clone` part.

```js
 if (React.Children.count(children) === 1 && typeof children === 'string') {		
    return <a href={href}>{children}</a>;		
￼ } else {		
￼    return children;
 }

```

WDYT?

alextrastero: the problem about this, is that if there's more than one child, ie icon and text it will not be a link anymore and remove functionality

alextrastero: that's why I though to just render all passed children

isaiah: Does the styling still work with the `<a>` wrapper?

alextrastero: I should think so yes

PR #354, opened 2017-09-01
Update OverlayTrigger.js

nsingal:
Modal options only work if the command is sent in with the options.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -31,7 +31,7 @@ class OverlayTrigger extends Overlay {
 
   showOverlay (e) {
     e.preventDefault();
-    $(`#${this.overlayID}`).modal(this.props.modalOptions).modal('open');
+    $(`#${this.overlayID}`).modal('open', this.props.modalOptions);
   }
 }
 


src/OverlayTrigger.js
```diff
@@ -31,7 +31,7 @@ class OverlayTrigger extends Overlay {
 
   showOverlay (e) {
     e.preventDefault();
-    $(`#${this.overlayID}`).modal(this.props.modalOptions).modal('open');
+    $(`#${this.overlayID}`).modal('open', this.props.modalOptions);
```
alextrastero: I'm not sure about this, it seems to work fine as it currently is, does it not work for you?

isaiah: he might mean there is not default value for `props.modalOptions`, and it doesn't work when the user didn't provide one?

alextrastero: ah I see, then maybe we could add `defaultProps: { modalOptions: {} }`?

isaiah: the way he proposed looks ok, why do we chain call `.modal`?

alextrastero: first we initialise and then we call it, but sure, I have no objection with the proposed solution. I just didn't know there was an initial issue. But now I see :)

alextrastero:
@nsingal did you have time to test this? The docs Modals don't work at the moment, if I revert this change they seem to work

PR #356, opened 2017-09-03
Add materialize-css as a peer dependency

alextrastero:


alextrastero:

PR #366, opened 2017-09-08
Init modals which have no trigger

cedretaber:
Hi, the modals of the sample page were not work, because which have not be initialized.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -12,11 +12,8 @@ class Modal extends Component {
   }
 
   componentDidMount () {
-    const { trigger, modalOptions } = this.props;
-
-    if (!trigger) {
-      $(`#${this.modalID}`).modal(modalOptions);
-    }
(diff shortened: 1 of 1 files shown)


alextrastero:
Just as a note, we need to initialise modals that contain a trigger and also when they don't contain a trigger

cedretaber:
I've clicked buttons but the modals of the examples didn't open. So, I considered that there were broken.

https://react-materialize.github.io/#/modals

It seems that the modals are never initialised which don't contain a trigger.

cedretaber:
Sorry, I have got a mistake.
It might not work only if it's on a sample page (because of react-live?)

alextrastero:
Yes! Once you change the code in react-live it doesn't get re-mounted so the initialise is not called. I'm not sure if we could maybe reinitialise on willReceiveProps ??

PR #373, opened 2017-09-13
1.0.13

alextrastero:
- Adds mocha plugin to eslint to resolve `.only` issues in specs
- Fixes Overlay default props issue
- Updates changelog
- Updates to new version

PR #377, opened 2017-09-19
Fixes OverlayTrigger

alextrastero:
Seems like setting defaultProps broke the OverlayTrigger, but setting the default value in a deconstructions works fine... javascript.


alextrastero:
Fixes by #373

PR #379, opened 2017-09-20
1.0.14

alextrastero:
Since 1.0.13 was already published I couldn't override, so this version is the next patch which also fixes broken Cards, please merge when reviewed


alextrastero:
cc @isaiah

Issue #388, opened 2017-10-02
Controlled input in Autocomplete


dmitry-yudakov:
Autocomplete component seems to support only uncontrolled input - the value is kept inside its internal state and I see no clear way of knowing when it's changed and also when the auto-complete  actually occurs.

I'll post a PR addressing this issue.

harishrohokale:
@alextrastero Is this fix available in latest release? I am also not able to get the value once autocomplete is done. Or atleast need an event which will tells us that autocomplete is done.

I tried onChange and onAutocomplete but both did not work when autocomplete is done

harishrohokale:
```<Autocomplete name='resource' onAutocomplete={this.onUpdateResource.bind(this)} title="Resource" s={12} data={this.RESOURCES}></Autocomplete>```

harishrohokale:
I have pulled down from └── react-materialize@1.0.16  (git://github.com/react-materialize/react-materialize.git#544429c5c3d1efa1317cd01fe08994d1951d4aa2)

alextrastero:
Did that make it work?

matas30rubliu:
@alextrastero id is hardcoded for Autocomplete = creating multiple Autocomplete fields creates them with the same ID, please allow it to set as a prop.

https://github.com/react-materialize/react-materialize/blob/master/src/Autocomplete.js#L103

alextrastero:
Please try `react-materialize3.x` and reopen if this is still happening, thanks!

PR #389, opened 2017-10-02
Autocomplete - controlled input (#388)

dmitry-yudakov:
This addresses lack of controlled input for Autocomplete component #388

### DIFF
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -1,3 +1,4 @@
+/* eslint no-unused-vars: ["error", { "ignoreRestSiblings": true }] */
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
@@ -9,12 +10,19 @@ class Autocomplete extends Component {
     super(props);
 
     this.state = {
-      value: ''
+      value: props.value || ''
     };
 
     this.renderIcon = this.renderIcon.bind(this);
     this.renderDropdown = this.renderDropdown.bind(this);
     this._onChange = this._onChange.bind(this);
+  }
+
+  componentWillReceiveProps (nextProps) {
+    const { value } = nextProps;
+    if (value !== undefined) {
+      this.setState({ value });
+    }
   }
 
   renderIcon (icon, iconClassName) {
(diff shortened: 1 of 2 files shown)


src/Autocomplete.js
```diff
@@ -9,14 +10,21 @@ class Autocomplete extends Component {
...
-      value: ''
+      value: props.value || ''
     };
 
     this.renderIcon = this.renderIcon.bind(this);
     this.renderDropdown = this.renderDropdown.bind(this);
     this._onChange = this._onChange.bind(this);
   }
 
+  componentWillReceiveProps (nextProps) {
```
alextrastero: You could even do `componentWillReceiveProps ({ value }) {`

alextrastero: p.s. doesn't the linter complain about space between parens and function name?

src/Autocomplete.js
```diff
@@ -136,7 +160,10 @@ Autocomplete.propTypes = {
   /**
    * Placeholder for input element
    * */
-  placeholder: PropTypes.string
+  placeholder: PropTypes.string,
+  onChange: PropTypes.func,
+  onAutocomplete: PropTypes.func,
+  value: PropTypes.string
 };
```
alextrastero: Please add a description to these fields so that they can be parsed and appear in the documentation, like L160 :)

src/Autocomplete.js
```diff
@@ -57,7 +65,19 @@ class Autocomplete extends Component {
...
+    const value = evt.target.value;
+    if (onChange) { onChange(evt, value); }
+
+    this.setState({ value });
+  }
+
+  _onAutocomplete (value, evt) {
+    const { onChange, onAutocomplete } = this.props;
+    if (onAutocomplete) { onAutocomplete(value); }
+    if (onChange) { onChange(evt, value); }
```
alextrastero: This method is for when you click on an element from the list right? why are we calling `onChange` also?

src/Autocomplete.js
dmitry-yudakov: Yep, I'll do it and I'll correct the parens

src/Autocomplete.js
dmitry-yudakov: I was wondering about it. Technically the value gets changed so we notify the user about it.
It will also allow us to rely only on onChange/value props and not use onAutocomplete if you don't care that it was autocompleted.

Otherwise you'd have to use both every time.

src/Autocomplete.js
dmitry-yudakov: Sure, I will

src/Autocomplete.js
dmitry-yudakov: Actually, it seems the parens are like this everywhere? 
The linter does complain when it's missing :)

alextrastero: Absolutely correct, I was confused with another project where I have the opposite settings :D

Issue #350, opened 2017-09-01
Select onChange getting fired twice


w3nisar:
Hi There,
  I am unable to figure out why a select onChange event triggers twice, please check the code

`import React, { Component } from 'react';
import '../css/Dashboard.css';
import { Preloader,Button,Icon,Modal,Row,Input } from 'react-materialize';

const Dashboard = React.createClass({

    checkResourceAccess(){
        alert("s");
    },

    render(){

        return (
            <div> 
                <Row>
                    <Input s={12} type='select' label="Select Transaction type" className="browser-default" onChange={this.checkResourceAccess}>
                      <option value="null">Select</option>
                      <option value="Select 1">Select 1</option>
                      <option value="Select 2">Select 2</option>
                      <option value="Select 3">Select 3</option>
                      
                    </Input>
                </Row>  
            </div>
        )
    }

});

export default Dashboard;`

alextrastero:
Looking into this :)

alextrastero:
Please try `react-materialize3.x` and reopen if this is still happening, thanks!

PR #391, opened 2017-10-03
fix for #350

jawadsh123:
This is a fix for #350 
Extracting onChange from "...other" seems to fix the issue.


alextrastero:
Hey @jawadsh123, thanks for the contribution! Would it be possible to add a spec for this? As in that the `onChange` is only called once? That would be awesome :)

alextrastero:

PR #412, opened 2017-10-26
Fix for #411

### LINKED ISSUE #411: An options array with a single element passed into a Select throws a "React.Children.only expected to receive a single React element child" error
If an options array of length one is passed as the children to a select input, a "React.Children.only expected to receive a single React element child" is thrown from the `Input.renderIcon` method.

This is a result of the `React.Children.count(children)` returning 1 when called on an array of length 1 and the `React.Children.only(children)` method expecting a single React element to be passed to it and not an Array.

I tried to write a fix by writing a check in `renderIcon` to test children was an Array and if so, pull the element out and pass it to `React.Children.only` (see below for the method). However this caused the text of the single option to be rendered next to the select box as an Icon. 

```
renderIcon () {
  const { icon, children } = this.props;
  if (icon) {
    return <Icon className='prefix'>{icon}</Icon>;
  } else {
    let icon = null;
    if (React.Children.count(children) === 1) {
      let _children = children;
      // If children is an Array of 1 element, pull the element out.
      if (Array.isArray(children) && children.length === 1) {
        _children = _children[0];
      }

      icon = React.Children.only(_children);
    }
    return icon === null ? null : React.cloneElement(icon, {className: 'prefix'});
  }
}
```

What is the use case for the `else` part of the method? 

I'll keep trying to work out a fix for it and create a merge request when I do however a bit more info on this method would be handy for debugging.

chris-gooley:
Fix #411 

Added check to test that children wasn't an array in `renderIcon`. See #411 for details.

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -242,7 +242,7 @@ class Input extends Component {
       return <Icon className='prefix'>{icon}</Icon>;
     } else {
       let icon = null;
-      if (React.Children.count(children) === 1) {
+      if (React.Children.count(children) === 1 && !Array.isArray(children)) {
         icon = React.Children.only(children);
       }
       return icon === null ? null : React.cloneElement(icon, {className: 'prefix'});
--- a/test/InputSpec.js
+++ b/test/InputSpec.js
@@ -125,6 +125,20 @@ describe('<Input />', () => {
         expect(wrapper.find('select').prop('multiple')).to.equal(true);
       });
     });
+
+    context('with only one option', () => {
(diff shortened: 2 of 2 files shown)


juanmanuelramallo:
I need this fix! Thanks @chris-gooley!

yuzhakovvv:
@alextrastero @chris-gooley it fails if child is an `ImmutableJS List` because check is only for `Array`

alextrastero:
Why not `toArray()` before passing to the component?

PR #413, opened 2017-10-26
TimePicker fix for examples 

chris-gooley:
The time picker in the example docs isn't working because the `type='time'` attribute was somehow removed/left out of `src/Input.js` when it was committed/merged in.

This just adds it back in.


chris-gooley:
My original comment was incorrect. Please ignore.

PR #430, opened 2017-11-28
Input value PropType can be array

### LINKED ISSUE #429: Input value PropTypes should allow for array values
When the Input component handles multiple values, the PropType for value should allow for PropType.array.

AleF83:
fixes #429

### DIFF
--- a/test/InputSpec.js
+++ b/test/InputSpec.js
@@ -22,6 +22,14 @@ describe('<Input />', () => {
     });
   });
 
+  context('#text with value', () => {
+    it('renders an input with value', () => {
+      const value = 'Some value';
+      const wrapper = shallow(<Input value={value} />);
+      expect(wrapper.find('input').prop('value')).to.equal(value);
+    });
+  });
+
   context('#text with label', () => {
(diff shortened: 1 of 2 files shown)


test/InputSpec.js
```diff
@@ -22,6 +22,14 @@ describe('<Input />', () => {
     });
   });
 
+  context('#text with value', () => {
+    it('renders an input with value', () => {
```
alextrastero: thanks for this, could we add another assertion for the `array` type?

alextrastero: like what we expect to appear when you add 2 values in the `value` prop

AleF83: I saw that there are another tests that check 'select' element that already validate it so I didn't add test for this case.

alextrastero: I think we should test that at least it doesn't break

AleF83: I've added test case

alextrastero:
Smashing! Thanks

alextrastero:
Released on `1.1.1`

Issue #447, opened 2017-12-23
Cannot read property 'props' of null 


sumansai14:
Passing null objects to Select Input children is giving an error. 
https://github.com/mui-org/material-ui/issues/8190

alextrastero:
Please try `react-materialize3.x` and reopen if this is still happening, thanks!

PR #448, opened 2017-12-23
allow invalid children for select inputs

sumansai14:
#447


alextrastero:

Issue #444, opened 2017-12-19
Add Prettier


alextrastero:
Reference implementation: https://github.com/styled-components/styled-components/pull/593

PR #451, opened 2018-01-07
Add prettier

basilbegonia:
https://github.com/react-materialize/react-materialize/issues/444

I hope this is a step in the right direction. This is my first PR to a real open-source project. Please let me know if there are any issues :)

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "semi": false,
+  "singleQuote": true
+}
--- a/package.json
+++ b/package.json
@@ -3,11 +3,7 @@
   "version": "1.1.2",
   "description": "Material design components for react",
   "main": "./lib/index.js",
-  "files": [
-    "LICENSE",
-    "README.md",
-    "lib"
-  ],
+  "files": ["LICENSE", "README.md", "lib"],
   "scripts": {
     "test": "mocha",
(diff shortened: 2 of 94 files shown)


alextrastero:
Hey @basilbegonia, first of all thanks for the contribution! please keep it up :)

I like prettier, I use it in some of my other projects, however for this one it seems like the default options deviate too much from our original `.eslintrc`. I'm up for adding some eslint --fix in a precommit hook but not prettier in this case.

Do you think we could adapt this PR?

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "semi": false,
```
alextrastero: Since the full project still uses semicolons, could we set this to true?

package.json
```diff
@@ -3,30 +3,23 @@
...
   "scripts": {
     "test": "mocha",
     "test:watch": "mocha --watch",
     "pretest": "npm run lint",
     "build": "npm test && babel -d lib src",
     "lint": "eslint src test; exit 0",
     "fix": "eslint src test --fix",
-    "release": "npm run build && npm publish"
+    "release": "npm run build && npm publish",
+    "eslint-check":
```
alextrastero: since this is not used, could we remove this task?

alextrastero:
No answer, closing this

PR #459, opened 2018-02-05
Added types

diabelb:
I just created index.d.ts file for typescript.


alextrastero:
Hey @diabelb, i'm not very used to Types, what is the purpose of this PR?

diabelb:
Types are necessary if you want to use react-materialize together with typescript. As typescript is strict typing language it needs d.ts files to determine what types components and component props are. If you don't find it usefull i can just pull it to DefinitelyTyped repository.

alextrastero:
Ah great! Would it be possible to add a section to the README? like using with Typescript???

diabelb:
Hi,

You just use it the same way as es6:
`import { Button, Col, Input, Row } from 'react-materialize';
`
Code editor (i use webstorm but others should works as well) will automatically use this definitions and will show components syntax.

Even if you are using es6 i recommend to you install @types for every packages that you using in your projects. It will work with es6 as well as typescript and help you writing code.

alextrastero:
So this PR is to help editors?

alextrastero:
Sorry I'm a types newbie

diabelb:
It will help editors and it's also required for typescript ;)

alextrastero:
I see, OK, do we need to configure anything else to use this lib with typescript? or is this it?

diabelb:
Configuration is the same as with es6. If you would like to test it quickly:
create new project using https://github.com/wmonk/create-react-app-typescript
configure materializecss, install react-materialize-ts from my npm repo and test some components. It should work same as with es6 ;)

maracuja-juice:
Wow. Thanks for the types. Will try it out in my local project.

Hope this will be merged soon.

alextrastero:
I would like to leave this here as a reference for anyone that needs info on this. We aren't looking into maintaining types for now :)

rlugge:
@diabelb if they aren't interested in adding types directly, any chance you coudl add this to definitely typed as @types/react-materialize?  I'd think about doing it myself, but it's honestly over my head.

Issue #468, opened 2018-02-25
DatePicker reopens when changing tabs.


shrestha-bikash:
Steps to reproduce this issue:

Click on the Date Picker input
Click on a date to choose one and click close or just click close
Click another tab
Return back to the previous tab
The Date Picker Modal opens without the user clicking on it

alextrastero:
Hey @shrestha-bikash, the code would look like this?
```jsx
<Tabs>
  <Tab>content</Tab>
  <Tab>
    <Input type='date' />
  </Tab>
  <Tab>content</Tab>
</Tabs>
```

Lucas-Braz-Cunha:
@alextrastero actually this happens if you just change tabs on the navigator (happened to me on Chromium).
```html
<Row>
  <Input type='date' />
</Row>
```

shrestha-bikash:
@alextrastero I am not talking about <Tabs> I am talking about browser tab like @Lucas-Braz-Cunha said.

alextrastero:
Ah! I see, would this be an issue with react-materialize or materialize-css then? We need to debug this, anyone want to give it a try?

shrestha-bikash:
I found that react-materialize has used pickadate [library](https://github.com/amsul/pickadate.js) and this issue is related to this library. [Here](https://github.com/amsul/pickadate.js/issues/160) is the issue that has the solution which can be used to solve this issue.

shrestha-bikash:
In Input.js file from src folder
here this code 
```
if (this.isDatePicker) {
    $(this.dateInput).pickadate(this.props.options);
    $(this.dateInput).on('change', this._onChange);
}
```
can be changed as
```
if (this.isDatePicker) {
    $(this.dateInput).pickadate({...this.props.options, onClose: function() {
       if(this.props.options.onClose) {
          this.props.options.onClose();
        }
        $(document.activeElement).blur()
      }
    });
    $(this.dateInput).on('change', this._onChange);
}
```

This may work I have done something like this to make it work in materialize-css.

alextrastero:
`react-materialize` is using whatever `materialize-css@0.98.2` was using.

It seems now they might have their [own implementation](https://github.com/Dogfalo/materialize/blob/7420a390d179900a399cd55cdd7d4f0fb8fce81a/js/datepicker.js)

alextrastero:
Great @shrestha-bikash! Wanna submit a PR?

shrestha-bikash:
Okay

[...]

PR #491, opened 2018-03-06
Fix datepicker reopening issue

shrestha-bikash:
# Description

The Input.js file from src folder has been changed. In `componentDidMount` while initializing pickadate there has been changes in the arguments. Additional `onClose` function is sent along with `options` where `onClose` function is called when user closes the date picker modal. This function blur the focused input element which causes the date picker modal reopening issue. The #468 issue has been solved with these changes. 
PS: This library is great for material design in react environment.

## Type of change

# How Has This Been Tested?

I tested this by building the repo and adding that built module to another project. It worked fine but I could not write the test case for this because there are lots of library like (sinon, chai) that are used which I am unfamiliar with. So I left this part to the other expert contributors.

# Checklist:

- [ ] My changes break one test case

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -25,13 +25,7 @@ class Input extends Component {
       $(this.selectInput).on('change', this._onChange);
     }
     if (this.isDatePicker) {
-      $(this.dateInput).pickadate({...this.props.options, onClose: function() {
-          if(this.props.options.onClose) {
-            this.props.options.onClose();
-          }
-          $(document.activeElement).blur()
-        }
-      });
+      $(this.dateInput).pickadate(this.props.options);
       $(this.dateInput).on('change', this._onChange);
     }
     if (this.isTimePicker) {
--- a/src/Input.js
+++ b/src/Input.js
@@ -25,7 +25,13 @@ class Input extends Component {
       $(this.selectInput).on('change', this._onChange);
     }
     if (this.isDatePicker) {
-      $(this.dateInput).pickadate(this.props.options);
+      $(this.dateInput).pickadate({...this.props.options, onClose: function() {
+          if(this.props.options.onClose) {
+            this.props.options.onClose();
+          }
+          $(document.activeElement).blur()
+        }
+      });
       $(this.dateInput).on('change', this._onChange);
     }
     if (this.isTimePicker) {
--- a/src/Input.js
+++ b/src/Input.js
@@ -25,7 +25,14 @@ class Input extends Component {
       $(this.selectInput).on('change', this._onChange);
     }
     if (this.isDatePicker) {
-      $(this.dateInput).pickadate(this.props.options);
+      let { options } = this.props;
+      $(this.dateInput).pickadate({...options,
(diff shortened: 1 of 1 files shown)


src/Input.js
```diff
@@ -25,7 +25,14 @@ class Input extends Component {
...
     }
     if (this.isDatePicker) {
-      $(this.dateInput).pickadate(this.props.options);
+      let { options } = this.props;
+      $(this.dateInput).pickadate({...options,
+        onClose: function () {
+          if (options.onClose) {
+            options.onClose();
+          }
+          $(document.activeElement).blur();
```
alextrastero: Nice! since this is not fully supported why don't we go with `$(this.dateInput).blur()` ?

alextrastero:
Also the specs would need a little update

alextrastero:
Build failing

src/Input.js
shrestha-bikash: Okay thats good.

shrestha-bikash:
The build is failing because test is failing for **renders a datepicker**. I haven't written any test for those changes because I don't have much knowledge about the libraries that are used for testing. I left that part to other people who has more knowledge about those libraries and testing.

alextrastero:
I think you can do this :)
https://github.com/react-materialize/react-materialize/blob/master/test/InputSpec.js#L248

Here we are expecting the materialize fn to be called with only `options`, but now, you are also passing a `onSomething` fn, so if you add an empty fn like "onXXX: () => {}" it should work.

Oh, and `npm run test:watch` to get the specs going :)

Let me know how it goes!

shrestha-bikash:
No its not working. I had tried those before sending PR. It didn't work so I left the testing part.

I tried this
```
const options = { one: 'two', onXXX: function () {} };
mount(<Input type='date' options={options} />);
expect(pickadateStub).to.have.been.calledWith(options);
```
I also tried this, its the same as above but still I tried just in case it may work.
```
 expect(pickadateStub).to.have.been.calledWith({...options, onXXX: function () {}});
```
I even tried this
```
expect(pickadateStub).to.have.been.calledWith(options, () => {});
```
I don't have any idea what this statement does. What type of constant does this make?
```
const pickadateStub = sinon.stub($.fn, 'pickadate');
```
If there are any other suggestions, please provide me I will try that.

src/Input.js
shrestha-bikash: `$(this.dateInput).blur()` this is not working because `this.dateInput` is undefined inside onClose function which is key value of an object.

alextrastero:
Hey @shrestha-bikash sorry for the late reply.
Try adding
`const options = { one: 'two', onClose: sinon.match.func };`
to the input spec.
This means that the test doesn't care about that function, only that it's a function.

The `const pickadateStub = sinon.stub($.fn, 'pickadate');` is to stub (fake) a method from an external library.
Have a quick look here: http://sinonjs.org/releases/v1.17.7/stubs/

alextrastero:

PR #494, opened 2018-03-06
Controlled modal

### LINKED ISSUE #493: Controlled modal component
### Description

Allow the modal component to be controlled via props without the need for a button. It would function similar to how material-ui opens/closes the modal via the `open` prop. http://www.material-ui.com/#/components/dialog

alextrastero:
# Description

Modal needs to be a controlled component, therefor it can accept `open` prop to be opened on mount.

Fixes #493 

## Type of change

# How Has This Been Tested?

With 'open' prop, expect modalStub to be called thrice.

# Checklist:

- [ ] I have not generated a new package version. (the maintainers will handle that)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.1.0
+
+- Adds `open` prop to Modal allowing it to 'opened' to mount
+
 ## 2.0.6
 
 - Updates to materialize-css@0.100.2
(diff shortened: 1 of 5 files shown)


severeon:
Shouldn't this logic be in the render function as well? Currently it will only open the modal via the prop if it was instantiated with the property set, since it is only checking for the property in componentDidMount

PR #495, opened 2018-03-06
Controlled modal

severeon:
# Description

Allow modal to open when the open property is set after initial component mount.  Added ability to close modal by setting the open prop to false.  This is a small fix to the feature added in #493

## Type of change

# How Has This Been Tested?

Added an additional test case to ModalSpec.js, and strengthened the existing test from #493 
verifies showModal is only called when the modal is no open by expecting three calls to stub.
tests that first call to $.modal receives modalOptions
tests that last call to $.modal stub passes 'open' or 'close' as only argument.

# Checklist:

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -34,6 +34,12 @@ class Modal extends Component {
     this.modalRoot = null;
   }
 
+  componentWillReceiveProps (nextProps) {
+    if (nextProps.open) {
+      this.showModal();
+    }
+  }
+
   renderModalPortal () {
     const {
       actions,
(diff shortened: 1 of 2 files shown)


src/Modal.js
```diff
@@ -74,6 +83,13 @@ class Modal extends Component {
     $(`#${this.modalID}`).modal('open');
   }
 
+  hideModal (e) {
+    if (e) e.preventDefault();
+    const { modalOptions = {} } = this.props;
+    $(`#${this.modalID}`).modal(modalOptions);
```
alextrastero: Do we need to initialise the modal again or would `close` be enough

severeon: We do not. I removed the call to $.modal and assert it isn't called an extra time when hiding the modal in the spec

alextrastero:
This is great stuff, thanks again for the contribution

PR #501, opened 2018-03-14
Initialize Tabs on new props

### LINKED ISSUE #498: Tab active not working
I have a header with links that direct to different pages using react-router. One of the pages has tabs on it, and when the page loads from being selected in the header, the component for the first tab (which I've set as active) is loaded, but the tab itself is not active. The problem here is that if I click on another tab (tab 2 for example), tab 2 becomes active, shows the content for that BUT also continues to show the content for tab 1. If I reload the page, the tabs work fine, but by navigating to the page via the header, I have this issue consistently.

screen shot 2018-03-14 at 10 01 21 am
screen shot 2018-03-14 at 10 01 30 am

alextrastero:
Fixes #498

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -21,6 +21,14 @@ class Tabs extends Component {
     if (onChange) onChange(idx, e);
   }
 
+  componentWillReceiveProps (nextProps) {
+    const { tabOptions = {} } = nextProps;
+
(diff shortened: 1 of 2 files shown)


Daramola98:
How do I use this I am having the same issue referenced in issue #498

PR #504, opened 2018-03-16
add container prop to Navbar and test classes.

hnoor:
# Description

added boolean container prop to navbar. 

## Type of change

# How Has This Been Tested?

Added test class and also using it in some of my other projects. 
# Checklist:


alextrastero:
Hey @hnoor, thanks for the contribution, instead of adding custom classes to the materialize components, why not pass a classNames prop that would be concatenated to `nav-wrapper`. I can't see any example in materialize-css with container class added:
http://materializecss.com/navbar.html

hnoor:
Hi @alextrastero .  

Navbar does have a classNames prop that passes css classes to the navbar. But when I tried adding the .container class that way this is what happened. 

original navbar with no container
`<Navbar brand={ Logo } right fixed className="grey lighten-4">`

no container prop

navbar with container passed withing the ClassNames prop
`<Navbar brand={ Logo } right fixed className="container grey lighten-4">`

container class

So I added a boolean container prop that works the same way as the fixed prop
`<Navbar brand={ Logo } right fixed container className="grey lighten-4">`

fixed container prop

alextrastero:
Thanks for the explanation @hnoor, however I still think that it's a very specific usecase, what react-materialize IS is just a react wrapper on top of materialize-css. If materialize-css doesn't have this feature, I would rather not add it.

If what you want is to pass a custom class to some element, let's move the className prop to the node which makes more sense.

alextrastero:
I think our issue is that className is passed to `nav`, let's move it to concatenate with `<div className='nav-wrapper'>` WDYT?

alextrastero:
Then again, would the color classes work then?

hnoor:
I thought of moving ClassName to concatenate with the nav-wrapper instead of adding to `<nav>`, but what I saw was that the color classes would only change within the wrapper, meaning that if you had a color and the container class you would end up with something like this

image

instead of this 

image

From what I understand, the only class that has to be added to the nav-wrapper is the container class because if you add the container to `<nav>` you get this 

image


Everything else works fine as it is.

alextrastero:
I see, ok. I think you have a valid point. Let's do this :)

src/Navbar.js
```diff
@@ -47,9 +48,15 @@ class Navbar extends Component {
...
+    let wrapper = 'nav-wrapper';
+
+    if (container) {
+      wrapper = 'nav-wrapper container';
+    }
+
     let content = (
       <nav {...other} className={className}>
-        <div className='nav-wrapper'>
+        <div className={wrapper}>
```
alextrastero: here let's use `cx`:
```jsx
<div className={cx('nav-wrapper', { container })}>
```

alextrastero: This basically means, add container class if container is true

src/Navbar.js
```diff
@@ -83,6 +90,11 @@ Navbar.propTypes = {
    * Makes the navbar fixed
    */
   fixed: PropTypes.bool,
+/**
+ * adds container to Navbar
```
alextrastero: `adds container class to nav-wrapper`

test/NavbarSpec.js
```diff
@@ -36,4 +36,9 @@ describe('<Navbar />', () => {
     wrapper = shallow(<Navbar fixed />);
     expect(wrapper.hasClass('navbar-fixed')).to.eq(true);
   });
+
+  it('can have container', () => {
+    wrapper = shallow(<Navbar container />);
+    expect(wrapper.find('nav-wrapper container')).to.have.length(0);
```
alextrastero: This should be have length 1 right? you can also use enzyme's `hasClass`

hnoor: `hasClass` here doesnt work.

`it('can have container', () => {
    wrapper = shallow(<Navbar container />);
    expect(wrapper.hasClass('container')).to.eq(true);
  });`

this returns `false` when `true` is expected

hnoor: has doesnt work if I use `mount`

test/NavbarSpec.js
```diff
@@ -38,7 +38,7 @@ describe('<Navbar />', () => {
   });
 
   it('can have container', () => {
-    wrapper = shallow(<Navbar container />);
-    expect(wrapper.find('nav-wrapper container')).to.have.length(0);
-  });
+    wrapper = mount(<Navbar container />);
+    expect(wrapper.props().container).to.eq(true);
```
alextrastero: We know the prop container is `true` because you are setting it on L41, let's check that hasClass container is true please :)

hnoor: yeah, you are right. check my recent changes. I think the issue is fixed now

alextrastero:
super

alextrastero:
Actually could you remove the package-lock file?

alextrastero:
Sorry @hnoor ! I meant that since your PR doesn't install any new packages you don't need to commit the `package-lock` file :)

hnoor:
@alextrastero Hi! We should be ready to close this request, right??

alextrastero:
Cool, closing then :)

Issue #465, opened 2018-02-13
Support `onClick` for <NavItem/>


lloydwatkin:
I'd like to be able to pass on `onClick` prop to `<NavItem/>` rather than a `href`.

alextrastero:
All props are passed here https://github.com/react-materialize/react-materialize/blob/master/src/NavItem.js#L12, doesn't this work?

lloydwatkin:
Unfortunately not, if I write:

```jsx
      <Navbar>
        <NavItemonClick={this.props.refreshAndBrowse}>
          <Icon left>view_list</Icon>
          <span>'BROWSE}</span>
        </NavItem>
      </Navbar>
```

This generates:

```html
<li class="no-padding">
  <a href=""><i class="material-icons left">view_list</i><span>Browse</span></a>
</li>
```

So when I click on the `<li/>` (to which the `onClick` prop is attached) the `<a/>` tag is getting activated (since it has a href) and hijacks the action.

lloydwatkin:
...if `href` attribute is not added to the `<a/>` tag then the `onClick` fires as expected and everything is good :+1:

lloydwatkin:
My short term solution is to use an `<li/>` in `<Navbar/>` rather than `<NavItem/>` but longer term (or before I get a chance to make a PR myself) it would be good to support the clicks only :)

alextrastero:
Ah I see, ok so ideally if prop `onClick` exists we should remove the `href` from `a`

lloydwatkin:
I'd say if `href` doesn't exist don't include it? That way its up to the
user?

marvinkome:
I think the way you should do this is to add `e.preventDefault()` to the onClick handler this way when the `a` is clicked it doesn't do what the browser default does that is route to the `href` location.

lloydwatkin:
That doesn't work. The link tag is a child element.

alextrastero:
@lloydwatkin @marvinkome feel free to give it a go :)

onluiz:
@alextrastero @lloydwatkin, this issue could be closed? :)

alextrastero:
Yes! :)

PR #517, opened 2018-04-01
add onClick suport for NavItem component #465

onluiz:
# Description

Added onClick suport for NavItem, suggested on issue #465. Also suggested by @alextrastero that href should be removed when onClick is present on NavItem, this way onClick will work without interference from href.

## Type of change

# How Has This Been Tested?

- Tested adding onClick only to NavItem. onClick should works normal
- Tested adding onClick and href to NavItem. onClick should works well and href is removed so it will not cause interference to onClick execution
- Tested adding only href to NavItem. Href should be executed normally
- Tested with no onClick nor href. No error shown

# Checklist:

### DIFF
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -9,9 +9,9 @@ const NavItem = ({
   ...props
 }) => {
   if (divider) return <li className='divider' />;
-  const a = onClick ? <a>{children}</a> : <a href={href}>{children}</a>;
+  const a = onClick ? <a onClick={onClick}>{children}</a> : <a href={href}>{children}</a>;
   return (
-    <li {...props} onClick={onClick}>
+    <li {...props}>
       {a}
     </li>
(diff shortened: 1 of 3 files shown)


src/NavItem.js
```diff
@@ -5,14 +5,14 @@ const NavItem = ({
...
 }) => {
   if (divider) return <li className='divider' />;
+  const a = onClick ? <a>{children}</a> : <a href={href}>{children}</a>;
   return (
-    <li {...props}>
-      <a href={href}>
-        { children }
-      </a>
+    <li {...props} onClick={onClick}>
+      {a}
```
alextrastero: why don't we add the `onClick` to the `a` tag?

onluiz: That would be simplier =)

I will adjust it.

PR #519, opened 2018-04-02
Adding prettier to project. Running prettier in all project

### LINKED ISSUE #444: Add Prettier

onluiz:
# Description

fix #444 

- Prettier added to project
- All files were formatted
- Each file changed will be formatted by prettier before each commit
- Rule of semi-column was kept

## Type of change

# How Has This Been Tested?

- Tested running npm test
- Tested running npm start in docs directory

# Checklist:


package.json
```diff
@@ -66,7 +67,16 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
+    "husky": "^0.14.3",
+    "lint-staged": "^7.0.2",
+    "prettier": "^1.11.1",
     "react": "^16.2.0",
```
alextrastero: these should be `devDependencies`

.eslintrc
```diff
@@ -3,13 +3,14 @@
 extends:
   - standard
   - standard-react
+  - prettier
```
alextrastero: do we need `standard` if we have `prettier`? I don't know if some/all rules are redundant

onluiz: I will remove standard and make some tests. But I think you're right.

package.json
onluiz: Done =)

onluiz:
@alextrastero, here is what I did:

- Adjusted prettier dependencies in package.json. Now they're into dev dependencies
- removed standard from eslitrc, keeping standard-react and prettier
- added eslint-plugin-prettier and eslint-config-prettier as devDependencies
- ran all tests a couple of times to make sure no new warning was shown

Could you review this? Want to make sure it is done the right way =)

Thank you very much!

alextrastero:
Thank you! Going to have a look asap

alextrastero:
`npm run lint` doesn't work, it's not using the rules that prettier uses

alextrastero:
Also now the fact that it's a commit hook you can't see code diff in the console

onluiz:
So, better remove commit book?

Also, let me fiz npm run lint.

alextrastero:
If it's ok with you I would rather keep single quotes and double for JSX

alextrastero:
So as far as I understand, now we have the choice to run `npm run prettier` and that would format our code

alextrastero:
I don't want to be a bummer but I think I would rather make `npm test` fail if `npm run lint` fails (so remove the `; exit 0`)

onluiz:
Haha that’s fine. Let me review it all and make another push.

We keep trying until it gets good. So please keep telling me what needs to be done. :)

onluiz:
> If it's ok with you I would rather keep single quotes and double for JSX
Done

> remove the ; exit 0
Done

alextrastero:
Perfect

Issue #416, opened 2017-10-29
Fixed (always visible) SideNav


MatheusVellone:
[Materialize SideNav docs](http://materializecss.com/side-nav.html#variations) (last example) tells about adding a `fixed` class in addition to `side-nav` to create a always visible SideNav.
Tried to accomplish this behavior with the lib, but couldn't.
I've searched docs, issues, SO questions, but didn't find anything.

Looks like this feature is missing in the lib.

alextrastero:
True, this is not implemented yet, adding a `help-wanted` label incase anyone want's to pick this up:
https://github.com/react-materialize/react-materialize/blob/master/src/SideNav.js
Adding fixed prop should add class to 'side-nav'

MatheusVellone:
I'm not too experienced with React, but I can work on a PR to fix it.

Just need some opinions (because I'm new with React)
If the `fixed` prop exists, then `trigger` shouldn't be required (doesn't need a `trigger` prop because the side-nav is always visible). So I was thinking about [this package](https://github.com/evcohen/react-proptype-conditional-require) to conditionally set the `required` rule for the `trigger` option. What do you think about it ?

alextrastero:
Hey @MatheusVellone, thank you! As for the package, I would rather not add another dependency but favour adding the prop `trigger` as optional instead of required, what do you say?

MatheusVellone:
IMO, I like the idea of `trigger` being required when `fixed` is false.
Because when someone try to use it without `fixed` and without `trigger`, at least a prop validation will be shown.

As this won't add/remove any functionality and add another dependency, if you're ok with it, I'll skip it.
Is skipped, I think in a message in the docs to warn possible incorrect usages is a good

Jiab77:
Hi all, I'll try to help on this as I'm using your in mine. Thanks for your great job, saved me a lot.

alextrastero:
Thanks @Jiab77

Jiab77:
@alextrastero You're welcome. Do you have some guidelines to modify your project ?

alextrastero:
@Jiab77 not really, we have some eslint rules, but that's it :)

alextrastero:
Hey @Jiab77 are you still up for this? :)

Jiab77:
Sorry, yes I'm still on, just having a work that taking me a lot of time. I took time to read the sourcecode, I've cloned your repository and started studying the way to include this features among other missing ones.
[...]

PR #520, opened 2018-04-03
Feature fixed sidenav

lyssareba:
# Description
This pull request is meant to allow the SideNav to be fixed as well as show the SideNav trigger on Large Displays. This should fix Issue #416  

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] This change requires a documentation update (Add showOnLarge to SideNav PropTypes
https://react-materialize.github.io/#/sidenav)

# How Has This Been Tested?

Using the example SideNav in the docs, I used create-react-app to build a test application. I imported and initiated the SideNav. After passing a fixed prop to SideNav, the navigation stayed fixed and the trigger is not visible. If a trigger is passed with fixed, it will display the trigger on mobile only. While passing the showOnLarge prop, I can now see the trigger on Large Displays as well as mobile. 

All tests were passing without adjustment. I added an extra test for the showOnLarge prop. 

I also removed the isRequired on trigger prop. The only problem here is, there is no error if you do not pass a trigger. To resolve this, I added a console.warn that lets the developer know there was no trigger passed to SideNav.

# Checklist:

### DIFF
--- a/src/SideNav.js
+++ b/src/SideNav.js
@@ -15,13 +15,10 @@ class SideNav extends Component {
   }
 
   render() {
-    const { className, children, ...props } = this.props;
+    const { className, children, trigger, fixed, ...props } = this.props;
     delete props.id;
-    delete props.trigger;
     delete props.options;
-    delete props.fixed;
-    const fixed = this.props.fixed || !this.props.trigger;
-    const classNames = cx('side-nav', { fixed }, className);
+    const classNames = cx('side-nav', { fixed: fixed || !trigger }, className);
 
     return (
       <span>
--- a/src/SideNav.js
+++ b/src/SideNav.js
@@ -22,6 +22,9 @@ class SideNav extends Component {
     delete props.showOnLarge;
 
     const classNames = cx('side-nav', { fixed }, className);
+    if (!this.props.trigger && !this.props.fixed) {
+      console.warn('No trigger was passed for the SideNav object. If no trigger was desired, please pass `fixed` to view your SideNav');
+    }
 
     return (
       <span>
--- a/src/SideNav.js
+++ b/src/SideNav.js
@@ -15,17 +15,17 @@ class SideNav extends Component {
   }
 
   render() {
-    const { className, children, fixed, ...props } = this.props;
+    const { className, children, ...props } = this.props;
     delete props.id;
     delete props.trigger;
     delete props.options;
-    delete props.showOnLarge;
-
+    delete props.fixed;
+    const fixed = this.props.fixed || !this.props.trigger;
     const classNames = cx('side-nav', { fixed }, className);
 
     return (
       <span>
-        {this.props.trigger && this.renderTrigger()}
+        {this.renderTrigger()}
         <ul id={this.id} className={classNames} {...props}>
           {children}
         </ul>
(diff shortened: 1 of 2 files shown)


alextrastero:
Also instead of a warning I would add more info to the proptypes, that way when creating the docs this will be parsed and added to the proptype table

lyssareba:
@alextrastero 
I added the showOnLarge prop just to make it easier for styling. It applies the hide-on-large-only if showOnLarge is not applied to maintain the materialize behavior. It also allows an easier flow from NavBar if you want to use NavBar on desktop and a SideNav on mobile.

I can remove those additions if you would prefer users just add the class names on the trigger. If that's the preferred method, it may be useful to include the common class names for those functionalities in an example. 

As for the proptypes, I'm not sure how to add the information you are referring too. Is there documentation or an example you could link? I'm only 6ish months into React.

alextrastero:
I see your point, the only issue is that we can't add a new prop for each scenario. Ideally we would have a generic solution that would work with other materialize-css helper classes. 
I do like however the idea that if there's no trigger the side nav should be fixed.

As for the docs, you need to add a comment in the propTypes section as other components have :)

https://github.com/react-materialize/react-materialize/blob/master/src/Carousel.js#L71

alextrastero:
By the way, I just merged a `prettier` PR so you might need to fixed some conflicts. Let me know if you need some help with that.

lyssareba:
My bad on the propTypes, I did add showOnLarge in my first commit in the propTypes.
```
  /**
   * Adds showOnLarge class to Trigger
   */
  showOnLarge: PropTypes.bool,
```

I didn't realize you were referring to those. 

I could adjust the SideNav to manage that logic itself. 

- If no trigger is passed, it automatically applies the fixed class.
- If a trigger is passed with fixed, it will apply the fixed class on desktop, but will hide the sideNav and display the trigger on med-and-down.
- If a trigger is passed without fixed, it will show on all screen sizes.

Is it too much to assume those use cases are the main issues?

alextrastero:
Let's give it a try! I would separate the trigger logic into a `renderTrigger` method so that it's easier to understand what's going on :)

Good luck!

lyssareba:
@alextrastero Everything should be done here.
Fixed prettier issues, and resolved that logic to not use showOnLarge. 
Shouldn't be any merge conflicts or test issues at this point.

src/SideNav.js
```diff
@@ -15,11 +15,12 @@ class SideNav extends Component {
   }
 
   render() {
-    const { className, children, fixed, ...props } = this.props;
+    const { className, children, ...props } = this.props;
     delete props.id;
     delete props.trigger;
     delete props.options;
-
+    delete props.fixed;
```
alextrastero: the way deconstruct works is that if you add `fixed` to L18 then you don't need to delete it from the `props` object

alextrastero: also add trigger to L18 and delete it from L20

src/SideNav.js
```diff
@@ -15,11 +15,12 @@ class SideNav extends Component {
...
+    const fixed = this.props.fixed || !this.props.trigger;
     const classNames = cx('side-nav', { fixed }, className);
```
alextrastero:
```jsx
const classNames = cx('side-nav', className, {
  fixed: fixed || !trigger
});
```

test/SideNavSpec.js
```diff
@@ -46,6 +46,11 @@ describe('<SideNav />', () => {
     assert.isTrue(sideNav.hasClass('side-nav fixed red'));
   });
 
+  it('should render trigger on large screens if `showOnLarge`', () => {
```
alextrastero: let's also add some specs to check the `fixed` logic and what happens when there's no `trigger`

src/SideNav.js
lyssareba: Didn't realize I could shorten this, this way. Thanks :)

Danny-G-Smith:
I would love to use this feature, however I am not sure where to place the props.  Best I can tell the documentation at https://react-materialize.github.io/#/sidenav didn't get updated.  Any help would be appreciated.

alextrastero:
Hey @Danny-G-Smith, to render a fixed sidenav just pass `fixed` as a prop ;)

PR #533, opened 2018-05-02
Section improvement

CloudPower97:
# Description

The Section component code is now refactored using ES6 classes syntax.
There is now the possibility to add custom classNames to this component and an id.
Moreover is now possible to change node. eg: <section> instead of <div>

## Type of change

# How Has This Been Tested?

I built the repo and tried this module in another project.

# Checklist:

### DIFF
--- a/src/Section.js
+++ b/src/Section.js
@@ -1,7 +1,11 @@
-import React from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
-const Section = props => <div className="section">{props.children}</div>;
+class Section extends Component {
+  render() {
+    return <div className="section">{this.props.children}</div>;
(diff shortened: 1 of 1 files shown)


src/Section.js
```diff
@@ -1,10 +1,33 @@
...
+  render() {
+    const { className, node, id, ...other } = this.props;
+
+    let C = node;
+    let classes = {
+      section: true
+    };
+
+    return (
+      <C {...other} id={id} className={cx(classes, className)}>
```
alextrastero: Since we are passing `...other` there's no real need to specify `id` is there?

CloudPower97: For some reason {...other} doesn't work, at least for me.

PR #543, opened 2018-05-06
Adding Icon dependency to chip component

CloudPower97:
# How Has This Been Tested?

Running `yarn test`

# Checklist:


alextrastero:
Same as before really, what are we trying to achieve here?

alextrastero:
And it's adding a dependency to `Icon`

CloudPower97:
@alextrastero Isn't adding the dependency to `Icon` a _good thing_? If in the future for some reason we change the rendering of icons the changes made will be also reflected here without the need to refactor this component.

alextrastero:
Good point, still for a small component like this we don't need a `class`, what if we just replace the `<i className` with an `Icon`?

CloudPower97:
@alextrastero right, going to do that ASAP.

CloudPower97:
@alextrastero Sorry to being so late

src/Chip.js
```diff
@@ -1,12 +1,13 @@
...
 import PropTypes from 'prop-types';
 import cx from 'classnames';
+import Icon from './Icon';
 
 const Chip = ({ children, close, className, ...props }) => {
   return (
     <div className={cx('chip', className)} {...props}>
       {children}
-      {close ? <i className="close material-icons">close</i> : null}
+      {close && <Icon className="close material-icons">close</Icon>}
```
alextrastero: Hey @CloudPower97, just realised that `<Icon/>` component already includes these classes
```suggestion
      {close && <Icon>close</Icon>}
```
would suffice

CloudPower97: Whoops, you are right, It was my oversight

CloudPower97:
@alextrastero ping

PR #545, opened 2018-05-06
consolidates main module root import/exports

tyler-reitz:
# Description

DRY's up top-level exports and allows dropping of unused modules in cases where tree-shaking is not supported or configured by importing components directly i.e. `import Button from "react-materialize/Button"`.

[Material-UI](https://github.com/mui-org/material-ui/blob/v1-beta/packages/material-ui/src/index.js), [Ant-Design](https://github.com/ant-design/ant-design/blob/master/components/index.tsx), and [React-Toolbox](https://github.com/react-toolbox/react-toolbox/blob/dev/components/index.js) all use this pattern in their top-level exports and Material-UI has even [written about it in detail in their docs](https://material-ui-next.com/guides/minimizing-bundle-size/).

Bundle-size reduction with minimal webpack config:

|                    | static             | parsed         | gzip                | bundle-analyzer |
| ------------- | --------------: | -------------: | --------------: | ------------------------------------- |
| __before__         | 280.87 KB    | 176.67 KB    | 42.93 KB        |   |
| __after__            | 133.23 KB    | 110.1 KB      | 34.72 KB        |  |
| __reduction__ |  52%             | 37%             | 19%               |  -- |

## Type of change

This wouldn't "break" existing functionality as the full API is still exposed at the top level for convenience. However, I did see an increase in final bundle size when importing the entire library using the current named import syntax i.e. `import { Button, Icon } from "react-materialize"` so marking as _breaking_ since there's a possible increase in bundle size.

# How Has This Been Tested?

I did not add a spec as there were none to begin with for the top-level module however the current test suite passes. I was able to manually test things locally by building and importing both directly from the components and top-level API.

# Checklist:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,87 +1,42 @@
-import Autocomplete from './Autocomplete';
-import Badge from './Badge';
-import Breadcrumb from './Breadcrumb';
-import Button from './Button';
-import Carousel from './Carousel';
-import Card from './Card';
-import CardPanel from './CardPanel';
-import CardTitle from './CardTitle';
-import Chip from './Chip';
-import Col from './Col';
-import Collapsible from './Collapsible';
-import CollapsibleItem from './CollapsibleItem';
-import Collection from './Collection';
-import CollectionItem from './CollectionItem';
-import Container from './Container';
-import Divider from './Divider';
-import Dropdown from './Dropdown';
-import Footer from './Footer';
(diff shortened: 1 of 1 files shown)


alextrastero:
I need to give this some thought :)

tyler-reitz:
For sure, no worries.

alextrastero:

PR #554, opened 2018-06-10
Including the option to change the active tab indicator color

brunogfranca:
Including the option to change the active tab indicator color via materializecss classes

# Description

Title self explanatory.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)
- [x ] This change requires a documentation update

# How Has This Been Tested?

Common change on projects using materializecss. Tested on local project and with npm test

# Checklist:

- [x ] My changes generate no new warnings
- [x ] I have not generated a new package version. (the maintainers will handle that)

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -34,6 +34,8 @@ class Tabs extends Component {
 
     const scope = `${idgen()}`;
 
+    const indicatorColor = `indicator ${this.props.indicatorColor}`
+
     return (
       <Row>
         <Col s={12}>
@@ -69,6 +71,7 @@ class Tabs extends Component {
                 </li>
               );
             })}
+            <div class={indicatorColor} style={{zIndex: 1}}></div>
           </ul>
         </Col>
         {React.Children.map(children, (child, id) => {
@@ -97,6 +100,7 @@ Tabs.propTypes = {
   className: PropTypes.string,
   defaultValue: PropTypes.string,
   onChange: PropTypes.func,
+  indicatorColor: PropTypes.string,
   /*
    * More info
    * <a href='http://materializecss.com/tabs.html'>http://materializecss.com/tabs.html</a>
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -34,7 +34,10 @@ class Tabs extends Component {
 
     const scope = `${idgen()}`;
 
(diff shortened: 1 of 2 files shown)


brunogfranca:
Found an error on the code and npm test is returning errors, closing PR until I fix it

src/Tabs.js
```diff
@@ -69,6 +74,7 @@ class Tabs extends Component {
                 </li>
               );
             })}
+            <div className={indicatorClasses} style={{ zIndex: 1 }} />
```
alextrastero: Hey @brunogfranca, this should be an `li` I think

src/Tabs.js
```diff
@@ -34,6 +34,11 @@ class Tabs extends Component {
 
     const scope = `${idgen()}`;
 
+    let indicatorClasses = 'indicator';
+    if (this.props.indicatorClasses) {
+      indicatorClasses += ` ${this.props.indicatorClasses}`;
```
alextrastero: let's deconstruct `indicatorClasses` on L33:
`const { children, className, defaultValue, indicatorClasses } = this.props;`

src/Tabs.js
alextrastero: We could also use here the `cx` method:
`<li className={cx('indicator', indicatorClasses)} style={{ zIndex: 1 }} />`

brunogfranca: I made it as a div (even though it's not proper html) because it was the popular fix to this issue. Haven't even consider testing if the li would work (just assumed it didn't). Tested it now and works just fine. I'll change it here 👍

src/Tabs.js
brunogfranca: Done 👍

brunogfranca:
You mean the color of the tab title? Totally agree

alextrastero:
So what about passing a prop to the `<Tabs color="red" />` and then using a lighter version of THAT color for `indicator`??

brunogfranca:
Perfect 👍 
I'll push it later today

alextrastero:
@brunogfranca need a hand with this one? :)

brunogfranca:
@alextrastero Sorry about the delay. I'm deploying a new project this week, so I'll probably take a look at this on the weekend =)

alextrastero:

PR #573, opened 2018-07-11
FAB Button tweaks

satom99:
Implements unhandled props for FAB buttons such as `disabled` and `tooltip*`.
In particular, an `id` has been added for a future implementation of [FeatureDiscovery](https://materializecss.com/feature-discovery.html).

_Note: I have not set a new package version._

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -59,16 +59,16 @@ class Button extends Component {
       classes['modal-' + modal] = true;
     }
     if (fab) {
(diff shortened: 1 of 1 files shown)


satom99:
As for having a separate method, I don't think it would make an improvement.
Mainly because the cleanness we gain we lose with variable handling.

And as for variable naming, absolutely! Let me know if the following changes suffice.

alextrastero:

PR #574, opened 2018-07-11
FeatureDiscovery

satom99:
Implements [FeatureDiscovery](https://materializecss.com/feature-discovery.html). Requires #573 to function.

_Note: I have not set a new package version._

### DIFF
--- a/examples/FeatureDiscovery.js
+++ b/examples/FeatureDiscovery.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import Button from '../src/Button';
+import FeatureDiscovery from '../src/FeatureDiscovery';
+
+export default
+<div>
+  <Button large floating fabClickOnly fab='horizontal' icon='edit' waves='light' className='red' id='menu'>
+    <Button floating icon='insert_chart' className='red'/>
+    <Button floating icon='format_quote' className='yellow darken-1'/>
+    <Button floating icon='publish' className='green'/>
+    <Button floating icon='attach_file' className='blue'/>
+  </Button>
+
+  <FeatureDiscovery open className='red white-text' target='menu'>
+    <h5>Title</h5>
+    <p>A bunch of text</p>
+  </FeatureDiscovery>
+</div>;
--- a/src/FeatureDiscovery.js
+++ b/src/FeatureDiscovery.js
@@ -38,9 +38,12 @@ class FeatureDiscovery extends Component {
 
     const parent = element.parent('.tap-target-wrapper');
     const origin = parent.find('.tap-target-origin');
-    const target = $("#" + this.props.target);
+    const target = $('#' + this.props.target);
     origin.click(
-      () => target.trigger('click')
+      () => {
+        target.trigger('click')
+        target.trigger('mouseover')
+      }
     )
   }
 
--- a/src/FeatureDiscovery.js
+++ b/src/FeatureDiscovery.js
@@ -1,0 +1,98 @@
+import React, { Component } from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+class FeatureDiscovery extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {mounted: false};
+    this.open = this.open.bind(this);
+    this.close = this.close.bind(this);
+    this.close = this.close.bind(this);
+    this.reference = this.reference.bind(this);
+  }
+
+  componentDidMount() {
+    this.setState({mounted: true});
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (!this.props.open && nextProps.open) {
+      this.open();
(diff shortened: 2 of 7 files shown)


alextrastero:
What's the diff with the previous PR? :)

alextrastero:
Wow, this is great, super stocked that this new feature is being developed. A few things are required for new features though, can we create a spec for this and also a example in `examples/`? 👏

satom99:
I added an example and a new page to the docs.
I am not entirely sure how we want the spec to be for this though :thinking:.

satom99:
Do note though that I've added a `mounted` bool. That-is so we make sure the button
that is set as the `target` is rendered first, and thus `$('#' + target)` succeeds.

The component will therefore be rendered and open once `componentDidMount` is called.
So in a sense yes; it will be open as soon as it is rendered.

examples/FeatureDiscovery.js
```diff
@@ -0,0 +1,18 @@
...
+export default
+<div>
+  <Button large floating fabClickOnly fab='horizontal' icon='edit' waves='light' className='red' id='menu'>
+    <Button floating icon='insert_chart' className='red'/>
+    <Button floating icon='format_quote' className='yellow darken-1'/>
+    <Button floating icon='publish' className='green'/>
+    <Button floating icon='attach_file' className='blue'/>
+  </Button>
+
+  <FeatureDiscovery open className='red white-text' target='menu'>
```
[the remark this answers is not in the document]
satom99: Do note though that I've added a `mounted` bool. That-is so we make sure the button
that is set as the `target` is rendered first, and thus `$('#' + target)` succeeds.

The component will therefore be rendered and open once `componentDidMount` is called.
So in a sense yes; it will be open as soon as it is rendered.

src/FeatureDiscovery.js
```diff
@@ -0,0 +1,101 @@
...
+    } else if (nextProps.open === false) {
+      this.close();
+    }
+  }
+
+  reference(element) {
+    this._element = element;
+    if (this.props.open) {
+      this.open();
+    }
```
alextrastero: I would rather move this logic to `componentDidMount` like other components do:
https://github.com/react-materialize/react-materialize/blob/188ee6c4521a1afc16548108cde2637d9bdc8ce6/src/Parallax.js#L7

alextrastero: Same as adding a ref to a node: https://github.com/react-materialize/react-materialize/blob/188ee6c4521a1afc16548108cde2637d9bdc8ce6/src/Tabs.js#L40

alextrastero: Then we wouldn't need this method `reference` which name is also misleading :)

src/FeatureDiscovery.js
```diff
@@ -0,0 +1,101 @@
...
+
+    const parent = element.parent('.tap-target-wrapper');
+    const origin = parent.find('.tap-target-origin');
+    const target = $('#' + this.props.target);
+    origin.click(
+      () => {
+        target.trigger('click')
+        target.trigger('mouseover')
+      }
+    )
```
alextrastero: Could you explain a bit what is happening here? reading the materialize-css docs it seems that you only need `element.tapTarget('open');` to show the featureDiscovery?

src/FeatureDiscovery.js
satom99: The reason why that's there is because the _tap target_ is not rendered until `mounted` is set to `true`.
This happens on `componentDidMount` because, as stated earlier, it's at this stage when we can assure
that the button we're targeting has been rendered already. Then, because `setState` is called, the component
is re-rendered and, because `mounted` is `true`, the actual _tap target_ is now rendered instead of an empty `div`.

Finally then we want to initialize the tap target accordingly once it's created, which is what happens within
the `reference` method.

Basically the issue here is that we cannot create the tap target before our target button is rendered/created
because the materialize library expects it to exist.

alextrastero: Could we solve this by passing the target as a prop, like with Modal?

alextrastero: Maybe an issue is that fab button never receives this ID
https://github.com/react-materialize/react-materialize/blob/master/src/Button.js#L85

src/FeatureDiscovery.js
satom99: This is here because when the _tap target_ is opened, a new button is created on top of the existing one
and no events are delegated to the original button. Which is essentially what we have there.

src/FeatureDiscovery.js
alextrastero: If I add a debugger in `componentDidMount` I can see that the target button is rendered

satom99: Could be a solution, I guess. I do like the idea of being able to specify an ID as the target though, instead of an element. We could perhaps support both?

alextrastero:

Issue #428, opened 2017-11-18
How can i use Toast component whitout Button


juan-carlos-correa:
I want to use Toast Component when i get the http response of a api and show the message.

Some like 

Materialize.toast('I am a toast!', 4000)

Just javascript code. Whithout button. Any idea?

alextrastero:
Hey @juan-carlos-correa, did you manage to solve this? Is there any improvements to the documentation you think is needed to solve this for others? Cheers!

dennisja:
call it as 
`window.Materialize.toast('I am a toast!', 4000)`

BrunaChamon:
@dennisja Solution works for me! Thank you!

juan-carlos-correa:
@alextrastero thanks, it was my mistake for my knowledge of js. The @dennisja solution is perfect ;D

ThatBrianDude:
Wow. The docs could really use a section of "programmatically doing xyz" for almost every component.

alextrastero:
PR's welcome :)

dennisja:
@merrazquin I'm sure you can remove it by doing something like this 
```js
const toast = document.querySelector('#toast-container>.toast.panning');
if(toast){
   toast.remove()
}
```

PR #576, opened 2018-07-13
Add docs for how to show and hide toasts programmatically.

hnrchrdl:
# Description

Read #428 and added some docs on how to show and hide a toast from programmatic context.

# Checklist:


examples/ToastProgrammatically.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import Button from '../src/Button'
+
+function() {
+    window.Materialize.toast('I am a toast!', 4000);
+}
```
alextrastero: what is the purpose of this function?

examples/ToastProgrammatically.js
```diff
@@ -0,0 +1,32 @@
...
+<div>
+    <Button onClick={() => {
+            window.Materialize.toast('I am a toast!', 50000);
+        }
+    }>
+        Click to show
+    </Button>
+    <Button onClick={() => {
+            const toast = document.querySelector('#toast-container>.toast');
+            console.log(toast)
```
alextrastero: leftover?

hnrchrdl: removed.

examples/ToastProgrammatically.js
hnrchrdl: removed.

hnrchrdl:
Thanks for reviewing and sorry for insufficiencies of my initial commits. Should be fixed now.
I will check the new beta when I find the time. Is it supposed to be mostly compatible with older versions? Glad to hear that JQuery is going to be dropped from this.

alextrastero:
Thanks for the updates @hnrchrdl.

As for the beta, it's in really early stages of development, it's a message to get users to write issues from real projects, thanks!

JazzBrown1:
Bit hacky but this works for me, for now:

Stuck it in a string and eval so my webpacks would compile it...... There are defo better ways of doing this...

  const showToast = (text) => {
    const string = `typeof M !== 'undefined' ? M.toast({html: '${text'}'}) : null;`;
    return eval(string);
  };

Issue #607, opened 2018-07-17
3.0.0 Tabs.js & Tab.js


alextrastero:
### Upgrade to latest materialize-css@1.0.0-rc.2

AC
- Branch off 3.0.0-beta.0
- Check that the component works
- Keep jQuery initialisation of component if exists
- Update with new available modifiers and options
- Update propTypes and description
- Update examples page
- Create PR into branch 3.0.0-beta.0

dmitry-yudakov:
I'd like to help with this

alextrastero:
Woohoo, yes :)

alextrastero:
Let me know if I can support you in anything

alextrastero:
Have a look at https://github.com/react-materialize/react-materialize/pull/613 if you need to

dmitry-yudakov:
#613 helped a lot, thanks :)

I think this issue is resolved and covered by tests update.

#408, #392 is fixed as well and manually tested but I think I can't cover it with enzyme test. Any advice for this?

I tried adding `expect(wrapper.find('#tab_10')).toHaveLength(1);` after _shallow/mount_ but it finds either 0 or 2 elements...

dmitry-yudakov:
Well, I was wondering if I should cover Tabs lifecycle refactoring by some test - so I tried to modify _Tabs.spec.js_ to check that elements ids are not changed anymore after re-render or properties update.

alextrastero:
You could add a spy to `idgen` and count how many times it was called?

dmitry-yudakov:
Yes, it worked, thanks

alextrastero:
Thanks for this @dmitry-yudakov

Issue #408, opened 2017-10-23
Change `Tabs` lifecycle initialise 


alextrastero:
Ref: https://github.com/react-materialize/react-materialize/issues/392

Currently the tabs are only initialised on mount, this should happen on receive props if they've changed.

Visovsiouk:
Can I try to resolve this issue?

alextrastero:
Of course @Visovsiouk :)

dmitry-yudakov:
@Visovsiouk Have you managed to make any progress here? I could try to address this issue in #607 for 3.0.0

alextrastero:
👍 @dmitry-yudakov

alextrastero:
Fixed in #634

PR #616, opened 2018-07-20
3.0.0 beta.0 tabs

dmitry-yudakov:
# Description

Issues #607, #408 are covered.

Adapted to new materialize-css API and refactored to be more efficient and fix wrong behavior after re-render.

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -7,6 +7,10 @@ import Row from './Row';
 import Col from './Col';
 
 class Tabs extends Component {
+  constructor(props) {
+    super(props);
+    this.scope = `${idgen()}`;
+  }
   componentDidMount() {
     const { tabOptions = {} } = this.props;
 
@@ -21,8 +25,8 @@ class Tabs extends Component {
     if (onChange) onChange(idx, e);
   }
 
-  componentWillReceiveProps(nextProps) {
-    const { tabOptions = {} } = nextProps;
+  componentDidUpdate() {
+    const { tabOptions = {} } = this.props;
 
     if (typeof $ !== 'undefined') {
       $(this._tabsEl).tabs(tabOptions);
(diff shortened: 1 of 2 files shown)


src/Tabs.js
```diff
@@ -21,25 +25,29 @@ class Tabs extends Component {
...
+    const { tabOptions = {} } = this.props;
 
     if (typeof $ !== 'undefined') {
       $(this._tabsEl).tabs(tabOptions);
     }
   }
 
+  componentWillUnmount() {
+    if (typeof $ !== 'undefined') {
+      console.log('destroy');
```
alextrastero: left over console.log

src/Tabs.js
```diff
@@ -21,25 +25,29 @@ class Tabs extends Component {
     if (onChange) onChange(idx, e);
   }
 
-  componentWillReceiveProps(nextProps) {
-    const { tabOptions = {} } = nextProps;
+  componentDidUpdate() {
```
alextrastero: any reason for changing this lifecycle?

src/Tabs.js
```diff
@@ -7,6 +7,10 @@ import Row from './Row';
 import Col from './Col';
 
 class Tabs extends Component {
+  constructor(props) {
+    super(props);
+    this.scope = `${idgen()}`;
```
alextrastero: 👍

alextrastero:
Try using mockClear in the specs

src/Tabs.js
dmitry-yudakov: Yes, it doesn't work correctly otherwise.

If you want to control active tab programmatically by passing 'active' property - $().tabs() should be called after it's rendered, while componentWillReceivedProps() will call it before - the tabs are out of sync after that.

Plus React are advising to use componentDidUpdate() and not use componentWillReceiveProps() anyway, so it seems like a good idea to change it.

dmitry-yudakov:
Yep, it's clearer this way, thanks

PR #619, opened 2018-07-23
 3.0.0 Icon.js

petergarnaes:
#595 I verified that Icons still work with version `1.0.0-rc.2` of the materialize style sheet, by creating a small project. Also added a small snapshot test to verify icons take placement as a prop. Slightly updated the documentation.

I know this is not a lot, but if anyone more creative than me can see anything that needs to be added, I will gladly give it a try :smile:


alextrastero:
Fixes #595

PR #630, opened 2018-08-11
Updated Slider for materialize 1.0.0 API #606

### LINKED ISSUE #606: 3.0.0 Slider.js & Slide.js
### Upgrade to latest materialize-css@1.0.0-rc.2

AC
- Branch off 3.0.0-beta.0
- Check that the component works
- Update with new available modifiers and options
- Update propTypes and description
- Update examples page
- Create PR into branch 3.0.0-beta.0

petergarnaes:
## Type of change

Fixes #606. API for this component has slightly changed. Renamed `transition` prop to `duration`, because it has been renamed in materialize 1.0.0. Added `height` prop which is passed along as option.

# How Has This Been Tested?

Tested like in PR #616 mocking the `M.Slider.init()` call. Added a test for fullscreen. There are some intricacies with `materialize-css` when updating the properties, which I have been unable to test since the test setup don't actually use `materialize-css`.

### DIFF
--- a/src/Slider.js
+++ b/src/Slider.js
@@ -6,6 +6,8 @@ class Slider extends Component {
   constructor() {
     super();
     this.initSlider = this.initSlider.bind(this);
+    this.fullscreenReset = this.fullscreenReset.bind(this);
+    this.setActiveIndex = this.setActiveIndex.bind(this);
   }
 
   initSlider() {
@@ -18,26 +20,36 @@ class Slider extends Component {
     });
   }
 
+  /**
+   * If the slider was not in fullscreen, the height is set as a style attribute
+   * on the Slider element. When `.destroy()` is called, this attribute is not
+   * removed, resulting in a fullscreen displayed incorrectly.
+   */
+  fullscreenReset(was_fullscreen) {
+    if (!was_fullscreen && this.props.fullscreen) {
+      this.instance.el.removeAttribute('style');
+      this.instance.el.childNodes[0].removeAttribute('style');
+    }
+  }
+
+  setActiveIndex(activeIndex) {
+    if (this.props.indicators && activeIndex)
+      this.instance['$indicators'][activeIndex].className =
+        'indicator-item active';
+  }
+
   componentDidMount() {
     this.initSlider();
   }
 
   componentDidUpdate(prevProps) {
-    if (this.instance) {
-      this.instance.destroy();
-      var activeIndex = this.instance.activeIndex;
-      // When instance is destroyed, the style attribute which determines the
-      // height of the slider is not removed, making the fullscreen incorrect height
-      if (prevProps.fullscreen === false && this.props.fullscreen) {
-        this.instance.el.removeAttribute('style');
-        this.instance.el.childNodes[0].removeAttribute('style');
-      }
-      this.initSlider();
-      // keep indicator at current index displayed
-      if (this.props.indicators && activeIndex)
-        this.instance['$indicators'][activeIndex].className =
-          'indicator-item active';
-    }
+    if (!this.instance) return;
+    const active_index = this.instance.activeIndex;
+    this.instance.destroy();
+    this.fullscreenReset(prevProps.fullscreen);
+    this.initSlider();
+    // keep indicator at current index displayed
+    this.setActiveIndex(active_index);
   }
 
   componentWillUnmount() {
--- a/src/Slider.js
+++ b/src/Slider.js
@@ -4,14 +4,19 @@ import cx from 'classnames';
 
 class Slider extends Component {
   componentDidMount() {
-    const { fullscreen, indicators, interval, transition } = this.props;
-
-    $('.slider').slider({
-      full_width: fullscreen,
+    const { indicators, interval, duration, height } = this.props;
+    this.instance = M.Slider.init(this._slider, {
       indicators,
       interval,
-      transition
+      duration,
+      height
     });
+  }
+
+  componentWillUnmount() {
+    if (this.instance) {
+      this.instance.destroy();
+    }
   }
 
   render() {
@@ -22,7 +27,10 @@ class Slider extends Component {
     };
 
     return (
-      <div className={cx(classes, className)}>
+      <div
+        ref={node => (this._slider = node)}
+        className={cx(classes, className)}
+      >
         <ul className="slides">{children}</ul>
       </div>
     );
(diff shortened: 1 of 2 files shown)


src/Slider.js
```diff
@@ -3,17 +3,49 @@ import PropTypes from 'prop-types';
...
+      height
     });
   }
 
+  componentDidMount() {
+    this.initSlider();
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.instance) {
```
onluiz: Hi! wouldn't it be nice if we use an earlier return here?

Like:
```
if(!this.instance) {
 return;
}
```

src/Slider.js
```diff
@@ -3,17 +3,49 @@ import PropTypes from 'prop-types';
...
+      this.instance.destroy();
+      var activeIndex = this.instance.activeIndex;
```
onluiz: Hey, my opinion :)

I would use `let` or `const` here.

And instead of using comments, I would create functions. This way the method would be smaller and more simple to ready.

What you think?

src/Slider.js
```diff
@@ -47,13 +82,20 @@ Slider.propTypes = {
...
-  transition: PropTypes.number
+  duration: PropTypes.number,
+  /**
+   * The height of the Slider window
+   * @default 400
+   */
+  height: PropTypes.number
 };
 
 Slider.defaultProps = {
```
onluiz: Do we have slider in the docs?
If not, wouldn't it be nice to put it there?
=)

petergarnaes: This is auto-generated, and put in the docs automatically. Very cool!

src/Slider.js
petergarnaes: You are absolutely right about `const` for `activeIndex`.

I guess functions are good to. However the functions would still need to be commented, to understand what is going on exactly.

onluiz: I see. =) Thanks @petergarnaes

src/Slider.js
```diff
@@ -3,17 +3,61 @@ import PropTypes from 'prop-types';
...
+    super();
+    this.initSlider = this.initSlider.bind(this);
+    this.fullscreenReset = this.fullscreenReset.bind(this);
+    this.setActiveIndex = this.setActiveIndex.bind(this);
+  }
 
-    $('.slider').slider({
-      full_width: fullscreen,
+  initSlider() {
+    const { indicators, interval, duration, height } = this.props;
```
alextrastero: In other components, we normally pass `options` prop, which is an object that is passed to initialise, please see https://github.com/react-materialize/react-materialize/blob/master/src/Carousel.js#L96 for an example

src/Slider.js
```diff
@@ -3,17 +3,61 @@ import PropTypes from 'prop-types';
...
+      height
     });
   }
 
+  /**
+   * If the slider was not in fullscreen, the height is set as a style attribute
+   * on the Slider element. When `.destroy()` is called, this attribute is not
+   * removed, resulting in a fullscreen displayed incorrectly.
+   */
+  fullscreenReset(was_fullscreen) {
```
alextrastero: nice catch! lets try and keep variables in the same case so `wasFullscreen` :)

src/Slider.js
```diff
@@ -3,17 +3,61 @@ import PropTypes from 'prop-types';
...
+        'indicator-item active';
+  }
+
+  componentDidMount() {
+    this.initSlider();
+  }
+
+  componentDidUpdate(prevProps) {
+    if (!this.instance) return;
+    const active_index = this.instance.activeIndex;
```
alextrastero: `activeIndex` please

src/Slider.js
```diff
@@ -3,17 +3,61 @@ import PropTypes from 'prop-types';
...
+    this.instance.destroy();
+    this.fullscreenReset(prevProps.fullscreen);
+    this.initSlider();
+    // keep indicator at current index displayed
+    this.setActiveIndex(active_index);
+  }
```
alextrastero: Just for documentation, could you explain in which scenario this would happen? As in, when would I update this component and want to keep the same `activeIndex`? :)

src/Slider.js
```diff
@@ -3,17 +3,61 @@ import PropTypes from 'prop-types';
 import cx from 'classnames';
 
 class Slider extends Component {
-  componentDidMount() {
-    const { fullscreen, indicators, interval, transition } = this.props;
+  constructor() {
+    super();
```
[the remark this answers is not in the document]
petergarnaes: Whoops!

src/Slider.js
petergarnaes: Whenever a react component get updated props, it should keep its state. Internally in the component it is not part of the state, but as a user, I would consider it so.

src/Slider.js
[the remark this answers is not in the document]
alextrastero: nono, it's not always needed, just a nice to have :)

src/Slider.js
alextrastero: Agree, thanks for the explanation

src/Slider.js
```diff
@@ -3,39 +3,37 @@ import PropTypes from 'prop-types';
...
       this.instance.el.childNodes[0].removeAttribute('style');
     }
   }
 
   setActiveIndex(activeIndex) {
-    if (this.props.indicators && activeIndex)
+    let { indicators } = this.props.options;
+    // In case this option is not defined, we assume true, as per default
+    let showIndicators = typeof indicators === 'undefined' || indicators;
+    if (showIndicators && activeIndex) {
```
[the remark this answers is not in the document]
petergarnaes: If `indicators` is undefined it should evaluate to true. `if (indicators && activeIndex)` if indicators is undefined, it evaluates to false.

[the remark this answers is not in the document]
onluiz: I see, but typeof indicators === 'undefined' || indicators says that "if indicators is undefined or not, we will do it"

if it is undefined or not will make no difference in this case cause it will always return true.

See if you agrees with me: https://codepen.io/onluiz/pen/KBjGYE?editors=1010

If not, please explain cause a may be wrong :nerd_face: 

Also, thanks for your help!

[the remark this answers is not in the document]
petergarnaes: The `indicators` variable is a boolean. So if `indicators = false` the expression is false.

petergarnaes:
I have demoed it during development, and seen it render correctly in the browser :+1:

PR #649, opened 2018-09-22
3.0.0-Carousel

CloudPower97:
# Description

This Pull Request is for #587.

Dropping support of `jQuery`.

Added two new options as per materialize-css documentation: 

- numVisible
- onCycleTo

Added an handy prop `centerImages` to make each image centered in the slider using `valign-wrapper` CSS helper.

## Type of change

# How Has This Been Tested?

`npm test` and in a side project.

# Checklist:

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -12,15 +12,29 @@ class Carousel extends React.Component {
   componentDidMount() {
     const { options = {} } = this.props;
 
-    if (typeof $ !== 'undefined') {
-      $(this._carousel).carousel(options);
+    if (typeof window.$ !== 'undefined') {
+      window.$(this._carousel).carousel(options);
+    } else if (typeof window.M !== 'undefined') {
+      this.instance = window.M.Carousel.init(this._carousel, options);
     }
   }
 
-  renderItems(child, idx) {
+  componentWillUnmount() {
+    if (typeof $ !== 'undefined') {
+      window.$(this._carousel).carousel('destroy');
+    } else if (typeof M !== 'undefined') {
+      this.instance.destroy();
+    }
+  }
+
+  renderItems(child, centerImages) {
     if (typeof child === 'string') {
       return (
-        <a className="carousel-item">
+        <a
+          className={cx('carousel-item', {
+            'valign-wrapper': centerImages
+          })}
+        >
           <img src={child} />
         </a>
       );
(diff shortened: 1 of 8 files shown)


src/Carousel.js
```diff
@@ -12,15 +12,29 @@ class Carousel extends React.Component {
   componentDidMount() {
     const { options = {} } = this.props;
 
+    if (typeof window.$ !== 'undefined') {
+      window.$(this._carousel).carousel(options);
+    } else if (typeof window.M !== 'undefined') {
```
alextrastero: let's just call it `M` as per the other implementations

CloudPower97: I'm having problem with this one. It doesn't work if I use "M" or "$" directly without using "window" before. I suspect it might be a problem of mine, though.
What should I do?

alextrastero: Are you having problems in the specs or when running locally?

CloudPower97: I'm having problem running the tests.
And, if I try to compile the library and use it in a project of mine it will not work as well.
The same problem I'm having on #650

alextrastero: Try it out on this repo: git@github.com:react-materialize/webpack-boilplate.git
Works fine for me :)

CloudPower97:
@alextrastero Not sure what is going on here... Still haven't figured out how to make tests not fail on my machine.

When I run `npm run test` ONLY tests for **Carousel** fails with `TypeError: Cannot read property 'init' of undefined`


No idea why these two are failing on Travis as well.
I can confirm everything works just fine when you use this component...

alextrastero:
Hey @CloudPower97, this could be a good initiation into jest mocking, have a look at how it's done here for reference:
https://github.com/react-materialize/react-materialize/pull/654/commits/c0fd8174a6501bcf0257b16b49b8c30361189231

Of course, if you don't have time let me know and I can give it a go :)

Hint:
```js
  const carouselInitMock = jest.fn();
  const carouselInstanceDestroyMock = jest.fn();
  const carouselMock = {
    init: (el, options) => {
      carouselInitMock(options);
      return {
        destroy: carouselInstanceDestroyMock
      };
    }
  };

  const restore = mocker('Carousel', carouselMock);
```

alextrastero:
Oh and as for Travis, it's failing because you removed the jQuery init but the specs are still expecting you to call the jQuery methods, which now need to be replaced with the new-mocker `M` methods.

CloudPower97:
@alextrastero Thanks for the cue

alextrastero:
Perfect @CloudPower97, really good job

PR #650, opened 2018-09-22
Supports both jQuery and JS initialization

CloudPower97:
# Description

Supports both jQuery and JS initialization, new options added (as per the[ documentation](https://materializecss.com/dropdown.html)), new rendering methods implemented.

Dropping `<NavItem>`s in favour of a more classical but flexible approach.

This was a necessity for me, as I'm using React Router.

With this change the integration with React Router is as simple as : 

```
<Dropdown trigger={<Button>Drop Me!</Button>}>
     <Link to="/first">First link</Link>
     <Link to="/second">Second link</Link>
</Dropdown>
```
 Note that dropping `<NavItem>`s  lead me to introducing `<Divider>` component to render dividers inside Dropdown;

If you want to render a divider, this is the code you should write: 

```
<Dropdown trigger={<Button>Drop Me!</Button>}>
      <Link to="/first">First link</Link>
      <Divider />
      <Link to="/second">Second link</Link>
</Dropdown>
```

I think this is more semantic as well.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

Added a test to check if rendering `<Divider />` is correctly implemented;

Every other tests are okay

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have not generated a new package version. (the maintainers will handle that)

PR #666, opened 2018-10-12
It's now possible to add extended components to the navbar

CloudPower97:
# Description

It's now possible to add extended components to the `NavBar`[ as per the doc](https://materializecss.com/navbar.html#navbar-tabs).
Also dropping _jQuery_ initialization and the use of `NavItem` component.
This was necessary to really make use of `React Router` `Link` component or any other custom component one may wish to use to fulfill routing purposes.

I think this approach is more flexible and resolve some edges cases, as in #650  

## Type of change

# How Has This Been Tested?

Used it in a project of mine and tested running `npm test`.
I've updated the snapshot accordingly to the changes.

Also added a new test to check if the mobile sidenav is rendered.

# Checklist:

### DIFF
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -1,13 +1,11 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-import Col from './Col';
 import Icon from './Icon';
 
 class Navbar extends Component {
   constructor(props) {
     super(props);
-    this.renderSideNav = this.renderSideNav.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


src/Navbar.js
```diff
@@ -1,91 +1,106 @@
...
 class Navbar extends Component {
   constructor(props) {
     super(props);
-    this.renderSideNav = this.renderSideNav.bind(this);
   }
 
   componentDidMount() {
-    if (typeof $ !== 'undefined') {
-      $('.button-collapse').sideNav(this.props.options);
+    const { options } = this.props;
```
alextrastero: could we set this default to `{}` just to not init with undefined?

src/Navbar.js
```diff
@@ -1,91 +1,106 @@
...
+    ));
 
-    let content = (
+    let navbar = (
       <nav {...other} className={className}>
         <div className="nav-wrapper">
-          <Col s={12}>
-            <a href={href} className={cx(brandClasses)}>
+          {brand && (
+            <a href="/" className={brandClasses}>
```
alextrastero: shouldn't we set this href to `href` prop?

src/Navbar.js
```diff
@@ -1,91 +1,106 @@
...
-              <Icon>view_headline</Icon>
-            </a>
-          </Col>
+          )}
+          <a href="#!" data-target="mobile-nav" className="sidenav-trigger">
+            <Icon>menu</Icon>
+          </a>
+          <ul className={navMobileCSS}>{links}</ul>
         </div>
+        {extendsWith && (
```
alextrastero: as per the [docs](https://materializecss.com/navbar.html) if we extend the navbar, it should be wrapped in an additional classname

src/Navbar.js
```diff
@@ -1,91 +1,106 @@
...
+      </React.Fragment>
+    );
   }
 }
 
 Navbar.propTypes = {
   brand: PropTypes.node,
   children: PropTypes.node,
   className: PropTypes.string,
+  extendsWith: PropTypes.arrayOf(PropTypes.node),
```
alextrastero: I would rename this prop to `extendWith` since extends is more for ES6 Class

test/Navbar.spec.js
```diff
@@ -42,6 +41,11 @@ describe('<Navbar />', () => {
   test('can be fixed', () => {
     wrapper = shallow(<Navbar fixed />);
     expect(wrapper).toMatchSnapshot();
-    expect(wrapper.hasClass('navbar-fixed')).toBeTruthy();
+    expect(wrapper.find('.navbar-fixed')).toBeTruthy();
```
alextrastero: We could assert that `toHaveLength(1)` to be more precise

alextrastero: same with other find -> true assertions

src/Navbar.js
```diff
@@ -1,91 +1,106 @@
...
+      navbar = <div className="navbar-fixed">{navbar}</div>;
     }
 
-    return content;
+    return (
+      <React.Fragment>
+        {navbar}
+
+        <ul
+          id="mobile-nav"
```
alextrastero: I would extract this ID into a constant like `MOBILE_NAV_ID`

src/Navbar.js
CloudPower97: Did that in `Navbar.defaultProps`

src/Navbar.js
CloudPower97: Right, my bad

CloudPower97:
@alextrastero ping :smile:

CloudPower97:
@alextrastero ping

Issue #599, opened 2018-07-17
3.0.0 Modal.js


alextrastero:
### Upgrade to latest materialize-css@1.0.0-rc.2

AC
- Branch off 3.0.0-beta.0
- Check that the component works
- Update with new available modifiers and options
- Update propTypes and description
- Update examples page
- Create PR into branch 3.0.0-beta.0

mendaomn:
According to [the migration guide](https://github.com/Dogfalo/materialize/blob/v1-dev/v1-upgrade-guide.md#modal)

> Modal
Change attribute data-activates to data-target
Removed ready and complete callback, use onOpenEnd and onOpenEnd callback instead

I've gone ahead and changed the propTypes accordingly. I'm not sure whether this name change should be added to our tests

All tests seem to be passing, is there another suggested way to "Check that the component works" ?

alextrastero:
Hey @mendaomn, that means manual QA, just fire it up locally and see if it works, is there a PR for this?

mendaomn:
I apologize for the very BASIC question, but what do you mean by "fire it up and see if it works"? The only way I found was running tests :)

alextrastero:
No problem!
If you clone this repo/branch: https://github.com/react-materialize/webpack-boilplate/tree/beta-debug on a folder next to your react-materialize repo and npm install; npm start you will be able to test your changes :)

PR #672, opened 2018-10-13
3.0.0-Modal

CloudPower97:
# Description

This Pull Request is for #599 .

Dropping __jQuery__ support.

Adding new options (as per the [documentation](https://materializecss.com/modals.html#options)):

- onOpenStart
- onOpenEnd
- onCloseStart
- onCloseEnd
- preventScrolling

Removed deprecated options:

-  ready
- complete

## Type of change

# How Has This Been Tested?

Updated spec to use the new mocker.

# Checklist:

### DIFF
--- a/test/Modal.spec.js
+++ b/test/Modal.spec.js
@@ -5,6 +5,8 @@ import mocker from './helper/new-mocker';
 
 describe('<Modal />', () => {
   let wrapper;
+  const modalOpenMock = jest.fn();
+  const modalCloseMock = jest.fn();
   const modalInitMock = jest.fn();
   const modalInstanceDestroyMock = jest.fn();
   const modalMock = {
@@ -13,7 +15,9 @@ describe('<Modal />', () => {
       return {
         destroy: modalInstanceDestroyMock
       };
-    }
+    },
+    open: modalOpenMock,
+    close: modalCloseMock
   };
   const restore = mocker('Modal', modalMock);
 
@@ -91,10 +95,8 @@ describe('<Modal />', () => {
   });
 
   // describe('controlled modal with `open` prop', () => {
-  //   let testModal;
-
   //   beforeEach(() => {
-  //     testModal = shallow(
+  //     wrapper = shallow(
   //       <Modal options={{ one: 1 }} open>
   //         {children}
   //       </Modal>
--- a/test/Modal.spec.js
+++ b/test/Modal.spec.js
@@ -1,12 +1,21 @@
 import React from 'react';
 import { shallow, mount } from 'enzyme';
 import Modal from '../src/Modal';
-import mocker from './helper/mocker';
+import mocker from './helper/new-mocker';
 
 describe('<Modal />', () => {
   let wrapper;
-  const modalMock = jest.fn();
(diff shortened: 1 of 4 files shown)


CloudPower97:
@alextrastero can you take a look at `controlled modal with `open` prop` test? It's driving me crazy because it says that `this.instance.open()` is not a function.

But it can't be right, as matter of fact it does indeed work when used in production AND `initializes with options` test simulate a click on a button which in turns call the `hide` method....

Surprise surprise it doesn't complain when about `this.instance.close()`.

alextrastero:
Will check ASAP

test/Modal.spec.js
```diff
@@ -1,12 +1,21 @@
...
+import mocker from './helper/new-mocker';
 
 describe('<Modal />', () => {
   let wrapper;
-  const modalMock = jest.fn();
-  const restore = mocker('modal', modalMock);
+  const modalInitMock = jest.fn();
+  const modalInstanceDestroyMock = jest.fn();
+  const modalMock = {
+    init: (el, options) => {
```
alextrastero: you would need to add here `open` method mock
```js
const modalOpenMock = jest.fn();
const modalMock = {
  open: modalOpenMock
  ...
```

CloudPower97: Hey @alextrastero, I've done that, but It still throws me the same error. Please, check the last commit

test/Modal.spec.js
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 import { shallow, mount } from 'enzyme';
 import Modal from '../src/Modal';
 import mocker from './helper/new-mocker';
+import idgen from '../src/idgen';
+
+jest.mock('../src/idgen');
```
alextrastero: i've mocked the `idgen` so that all modals have the same ID

test/Modal.spec.js
```diff
@@ -13,16 +16,18 @@ describe('<Modal />', () => {
     init: (el, options) => {
       modalInitMock(options);
       return {
+        open: modalOpenMock,
```
alextrastero: open and close are the methods returned by `init` :)

test/Modal.spec.js
```diff
@@ -65,7 +69,6 @@ describe('<Modal />', () => {
 
     afterEach(() => {
       modalInitMock.mockClear();
-      document.body.removeChild(document.body.lastElementChild);
```
alextrastero: this shouldn't be needed since the component should auto clean itself

test/__snapshots__/Modal.spec.js.snap
```diff
@@ -1,5 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`<Modal /> controlled modal with \`open\` prop mounts opened 1`] = `
+<div>
+  <Portal />
+</div>
```
alextrastero: this snapshot is updated since it's now shallowed

test/Modal.spec.js
CloudPower97: Ah, I see... Too bad I didn't catch this one! :grimacing:

PR #678, opened 2018-10-14
3.0.0-Chip

CloudPower97:
# Description

This Pull Request is for #588 

It's now possible to use the JavaScript Plugin.

These are the new options added (as per the [documentation](https://materializecss.com/chips.html#options)):

- data
- placeholder
- secondaryPlaceholder
- autocompleteOptions
- limit
- onChipAdd
- onChipSelect
- onChipDelete

## Type of change

# How Has This Been Tested?

I have added a new test using the `new-mocker` helper.

# Checklist:

### DIFF
--- a/src/Chip.js
+++ b/src/Chip.js
@@ -17,7 +17,8 @@ class Chip extends Component {
   }
 
   render() {
-    const { children, close, className, options } = this.props;
+    const { children, close, className, options, ...other } = this.props;
+
     const classes = cx(
       {
         chip: !options,
@@ -36,10 +37,12 @@ class Chip extends Component {
       chipContent = null;
     }
 
+    delete other.close;
+
     return (
       <div
         className={classes}
-        {...this.props}
+        {...this.other}
         ref={div => {
           this._chips = div;
         }}
--- a/src/Chip.js
+++ b/src/Chip.js
@@ -1,12 +1,13 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
+import Icon from './Icon';
 
 class Chip extends Component {
   componentDidMount() {
     const { options } = this.props;
 
-    if (options && M) {
+    if (options && typeof M !== 'undefined') {
       this.instance = M.Chips.init(this._chips, options);
     }
   }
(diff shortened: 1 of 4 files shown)


src/Chip.js
```diff
@@ -1,23 +1,95 @@
...
-      {children}
-      {close ? <i className="close material-icons">close</i> : null}
-    </div>
-  );
-};
+class Chip extends Component {
+  componentDidMount() {
+    const { options } = this.props;
+
+    if (options && M) {
```
alextrastero: if `options` is not defined, this component shouldn't initialise?

src/Chip.js
```diff
@@ -1,23 +1,95 @@
...
+      {
+        chip: !options,
+        chips: options
+      },
+      className
+    );
+    let chipContent = (
+      <React.Fragment>
+        {children}
+        {close ? <i className="close material-icons">close</i> : null}{' '}
```
alextrastero: should we add here `<Icon>` component like in others?

alextrastero: what's the empty string for?

src/Chip.js
```diff
@@ -1,23 +1,95 @@
...
+    );
+
+    if (options) {
+      chipContent = null;
+    }
+
+    return (
+      <div
+        className={classes}
+        {...this.props}
```
alextrastero: Don't we get a warning for unwanted props on `div` like this?

src/Chip.js
CloudPower97: Nope, as matter of fact you may want to use a `Chip` but not the __[JavaScript plugin](https://materializecss.com/chips.html#basic)__.
Also, the __CSS__ is different too, `chip` when you do not want to use the plugin, `chips` when you want to.

src/Chip.js
CloudPower97: > what's the empty string for?

Whoops, my bad!

src/Chip.js
alextrastero: Cool, ok

CloudPower97:
@alextrastero ping :smile:

alextrastero:
Check out my latest commits @CloudPower97 :)

Issue #642, opened 2018-08-28
Autocomplete options cannot be selected with arrow keys and enter


Biboswan:
### Steps to Reproduce
1. Follow html setup given in readme
2. Use the `Autocomplete` component with some data prop. Let have some data keys with same initials.
 Eg : ```data: {
        "Apple": null,
        "Aot":null,
        "Aoto":null,
        "Microsoft": null,
        "Google": 'https://placehold.it/250x250',
        "Ap":null
      }  ```

**Expected behavior:** 
Use of up/down arrows keys and enter to select an option. 

**Actual behavior:** 
Arrows cannot be used to move cursor and select an option. Mouse is the only way to select
### Versions

**react-materialize:** 2.4.5
**materialize-css:** 0.100.2
**react:** 16.4.2

PR #690, opened 2018-11-06
Add arrow support

mustaphaturhan:
# Description

Added arrow navigation and change with enter to Autocomplete. This PR is related with #642. 

## Type of change

# How Has This Been Tested?

I tried <Autocomplete /> with this object: `data: {
      Apple: null,
      Aot: null,
      Aoto: null,
      Microsoft: null,
      Google: 'https://placehold.it/250x250',
      Ap: null
    };`

When you press arrow down or arrow up, code is adding "active" class to li. When you press "ESC" key, value's state clears itself and "autocomplete dropdown" is closing. When you press enter, code searching value in "data" object and when find it; calling this._onAutocomplete function.

# Checklist:

### DIFF
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -39,10 +39,8 @@ class Autocomplete extends Component {
       return null;
     }
 
-    let matches = Object.keys(data).filter(key => {
-      const index = key.toUpperCase().indexOf(value.toUpperCase());
-      return index !== -1 && value.length < key.length;
-    });
+    let matches = this._findRealValue(value);
+
     if (limit) matches = matches.slice(0, limit);
     if (matches.length === 0) {
       return null;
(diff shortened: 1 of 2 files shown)


src/Autocomplete.js
```diff
@@ -79,7 +87,58 @@ class Autocomplete extends Component {
...
+        });
+      }
+    }
+
+    // if event is esc, reset value.
+    if (evt.keyCode === 27) {
+      this.setState({ value: '' });
+    }
+
+    if (evt.keyCode === 13) {
```
alextrastero: let's add a comment above this line to be consistent :)

src/Autocomplete.js
```diff
@@ -79,7 +87,58 @@ class Autocomplete extends Component {
...
+      evt.preventDefault();
+      const liValue = matches[activeItem].textContent;
+      let value = Object.keys(this.props.data).filter(key => {
```
alextrastero: prefer `const`

src/Autocomplete.js
```diff
@@ -79,7 +87,58 @@ class Autocomplete extends Component {
...
+        const index = key.toUpperCase().indexOf(liValue.toUpperCase());
+        return index !== -1 && liValue.length <= key.length;
+      });
```
alextrastero: this chunk is duplicated on L41, can we extract this to a method?

alextrastero:
And thanks for the contribution! keep it up

odunluzikkim:
> And thanks for the contribution! keep it up

Thank you for remarks.  I changed and pushed everything that you said and I added more comments to `evt.keyCode === 13` because it seems like little complicated.

mustaphaturhan:
ping @alextrastero

victorreinor:
When i select the option of the Autocomplete component using the enter key, the return of value is a Array, but when i select the option using the mouse click the return is a String, in this way i have a problem on my application because each time the return will be of a type. I recomend of the return always be a string how in the old version.

alextrastero:
Aha, good catch @victorreinor, is this easy to solve @mustaphaturhan ?

victorreinor:
After post the comment with the error, i try find a solution, i solved the problem puting a convertion .toString(); in the value variable inside the function _onAutocomplete in the file Autocomplete.js.
follow below the image with my changes. I hope thats help, thank you.
image

mustaphaturhan:
I am working on it @victorreinor, @alextrastero.

PR #695, opened 2018-11-21
Autocomplete

mustaphaturhan:
# Description

There was an error related with https://github.com/react-materialize/react-materialize/pull/690#issuecomment-440378694

Autocomplete showed multiple results when there is a more than one result. So, I fixed that problem. Users can't press return if there are no results anymore.

## Type of change

# How Has This Been Tested?

I tried it with:

```
const data = {
  Apple: null,
  Aot: null,
  Aoto: null,
  Microsoft: null,
  Google: 'https://placehold.it/250x250',
  Ap: null,
  Cd: null,
  Cdd: null,
  Cddrf: null
};

<Autocomplete data={data} limit={4} />
```

# Checklist:

### DIFF
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -39,7 +39,7 @@ class Autocomplete extends Component {
       return null;
     }
 
-    let matches = this._findRealValue(value);
+    let matches = this._findMatches(value);
 
     if (limit) matches = matches.slice(0, limit);
     if (matches.length === 0) {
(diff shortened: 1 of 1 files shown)


mustaphaturhan:
ping @alextrastero

alextrastero:
Thanks @mustaphaturhan, could this have been caught with a spec? Is there a spec we could add to test this?

mustaphaturhan:
Sorry, I don't know how to write tests but I can describe what was the problem before this fix @alextrastero. 

If you go to https://github.com/react-materialize/react-materialize/pull/690#issuecomment-440378694 and look for the problem, you can see an array problem as described. I made some research and found my mistake.

I thought my "findRealValue" function selects right one, but I was wrong. My function didn't select right match, it just creates an array that contains all of matched words. So, my function returned "Cd, Cdd, Cddrf" when you write "Cd" to input and select one item with pressing enter key. Input value was return as "Cd, Cdd, Cddrf".

Also, I have to give one more example because of my bad English i think. Let's think about "Aot, Aoto" values. When you write "Ao" to your input and select one with pressing enter key; input value was return as "Aot, Aoto." It's wrong because input value should return as selected item.

With this fix, input value returns Aot if you select Aot. And It isn't array anymore, it is string.

alextrastero:
Hey @mustaphaturhan, I'm afraid this is no longer needed, we decided to revert the arrow navigation PR since this feature is added natively by materializecss in the next build we are working on.

More info here: https://github.com/react-materialize/react-materialize/pull/704

PR #699, opened 2018-11-23
Parallax component enhancement

CloudPower97:
# Description

During development on a side project of mine I noticed that it was impossible to use `style` props or `id` 
on the `Parallax` component.

Moreover, looking at the source code I noticed that the `PropTypes` definition for options was completely wrong and the default options for the `materialize` plugin was just an empty object, instead of the default one as described by the [`materialize` documentation](https://materializecss.com/parallax.html#options).

I also took the occasion to make it possible to pass children to this component, as this was a thing I was experimenting with and I'm pretty happy with the result.

## Type of change

# How Has This Been Tested?

I've added a test to ensure that if no `options` are given to this component it would initialize itself with the default one as described by the [documentation](https://materializecss.com/parallax.html#options).

Also added a test to check if the it can correctly render a `children`, as a title, an icon or whatever you want.

# Checklist:

### DIFF
--- a/test/Parallax.spec.js
+++ b/test/Parallax.spec.js
@@ -36,11 +36,28 @@ describe('<Parallax />', () => {
       expect(parallaxInitMock).toHaveBeenCalledTimes(1);
     });
 
+    test('should have default options if none are given', () => {
+      mount(<Parallax />);
+      expect(parallaxInitMock).toHaveBeenCalledWith({
+        responsiveThreshold: 0
+      });
+    });
+
     test('should call Parallax with the given options', () => {
       const options = { responsiveThreshold: 200 };
 
       mount(<Parallax options={options} />);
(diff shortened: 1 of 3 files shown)


CloudPower97:
@alextrastero ping :smile:

test/Parallax.spec.js
```diff
@@ -36,11 +36,28 @@ describe('<Parallax />', () => {
...
+    });
+
     test('should call Parallax with the given options', () => {
       const options = { responsiveThreshold: 200 };
 
       mount(<Parallax options={options} />);
       expect(parallaxInitMock).toHaveBeenCalledWith(options);
     });
+
+    test('can render children element', () => {
```
alextrastero: how about we add the required prop and use to match snapshot here?

CloudPower97: @alextrastero please, have a look whenever you can :grinning:

PR #700, opened 2018-11-23
Carousel enhancement

CloudPower97:
# Description

Looking at the source code I noticed that when no options are given the `Carousel` component used just an empty object.

Hence I've added the default options as per the [`materialize` documentation](https://materializecss.com/carousel.html#options)

For accessibility purposes every image has now a default empty `alt` tag (also called the NULL alt text), implying they are just decorative images that serves no specific purpose.

I've also fixed a bug regarding `centerImages` prop introduced by #649, it was used only when using the `images` props.

It didn't work for `children` elements.

## Type of change

# How Has This Been Tested?

Added a test to check if the `centerImages` props is correctly used.
Added a test to make sure to initializes the component with the default options if none are given. (Refer to the [`materialize`](https://materializecss.com/carousel.html#options) documentation for more information)

# Checklist:

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -82,20 +82,16 @@ class Carousel extends React.Component {
 
 Carousel.propTypes = {
   /*
-  * Children to render as slider elements
+  * Children to render as carousel elements
   */
   children: PropTypes.any,
   /*
   * Array of image url's
   */
-  images: PropTypes.arrayOf(
-    PropTypes.shape({
-      src: PropTypes.string.isRequired,
-      alt: PropTypes.string
-    })
-  ) /*
+  images: PropTypes.arrayOf(PropTypes.string),
+  /*
   * Makes the images centered inside the carousel using '.valign-wrapper' CSS helper
-  */,
+  */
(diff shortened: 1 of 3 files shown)


CloudPower97:
@alextrastero ping

src/Carousel.js
```diff
@@ -29,30 +29,31 @@ class Carousel extends React.Component {
...
   render() {
     const {
       children,
       className,
       carouselId,
+      fixedItem,
       images,
       centerImages,
-      options = {}
+      options: { fullWidth }
```
alextrastero: this is already a deconstruction, lets use options to avoid confusion:)

src/Carousel.js
```diff
@@ -65,11 +66,11 @@ class Carousel extends React.Component {
           }}
           className={cx(
             'carousel',
-            { 'carousel-slider': options.fullWidth },
+            { 'carousel-slider': fullWidth },
```
alextrastero: I would keep this as options.fullwidth

CloudPower97: @alextrastero fixed :+1:

PR #701, opened 2018-11-23
Slide component enhancement

CloudPower97:
# Description

Refactoring the `Slide` component into a functional component, adding a `Caption` component.

This leads to a much more semantic and flexible way to define complex captions.

For example you can now add more component to a `Caption`, like a `Button`:
```javascript
<Slider className="white" options={options}>
     <Slide image={FirstSlide}>
          <Caption placement="left">
               <h4>This is our big Tagline!</h4>
               <h5 className="light grey-text text-lighten-3">
                     Here&apos;s our small slogan.
                </h5>
                <Link to="/whatever">
                     <Button className="hoverable" large>
                          Read More
                     </Button>
                </Link>
           </Caption>
    </Slide>
</Slider>
```
You can now use a [`picture`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) html element for [art direction](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction) or simply to provide different version of your images (WebP images, JPEG...)

For example 

```javascript

import {
     firstSlideWebP,
     firstSlideJpg,
} from '../../assets/img'

const FirstSlide = (
  <picture>
    <source srcSet={firstSlideWebP} type="image/webp" />
    <source srcSet={firstSlideJpg} type="image/jpg" />
    <img src={firstSlideJpg} alt="" />
  </picture>
)

<Slider className="white" options={options}>
     <Slide image={FirstSlide}>
          <Caption placement="left">
               <h4>This is our big Tagline!</h4>
               <h5 className="light grey-text text-lighten-3">
                     Here&apos;s our small slogan.
                </h5>
                <Link to="/whatever">
                     <Button className="hoverable" large>
                          Read More
                     </Button>
                </Link>
           </Caption>
    </Slide>
</Slider>
```

P.S: As noted also on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#Usage_notes) you might need to use `object-fit` and/or `object-position` to correctly display the `img`.
 
P.P.S: You may wish to take a look at [react-responsive-picture](https://github.com/braposo/react-responsive-picture) for the last use case

## Type of change

# How Has This Been Tested?

Add test for `Caption` element.
Updated every snapshots.

# Checklist:

### DIFF
--- a/src/Caption.js
+++ b/src/Caption.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import constants from './constants';
+
+const Caption = ({ className, placement, children, ...props }) => {
+  return (
+    <div
+      className={cx(
+        'caption',
+        {
+          [`${placement}-align`]: placement
+        },
+        className
+      )}
+      {...props}
+    >
+      {children}
+    </div>
+  );
+};
+
+Caption.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  /**
+   * Placement of the caption
+   * @default 'center'
+   */
+  placement: PropTypes.oneOf(constants.PLACEMENTS),
+  style: PropTypes.object
+};
+
+Caption.propDefault = {
+  placement: 'center'
+};
+
+export default Caption;
--- a/src/Caption.js
+++ b/src/Caption.js
@@ -5,16 +5,7 @@ import constants from './constants';
 
 const Caption = ({ className, placement, children, ...props }) => {
   return (
-    <div
-      className={cx(
-        'caption',
-        {
-          [`${placement}-align`]: placement
-        },
-        className
-      )}
-      {...props}
-    >
+    <div className={cx('caption', `${placement}-align`, className)} {...props}>
       {children}
     </div>
   );
@@ -27,11 +18,10 @@ Caption.propTypes = {
    * Placement of the caption
    * @default 'center'
    */
-  placement: PropTypes.oneOf(constants.PLACEMENTS),
-  style: PropTypes.object
+  placement: PropTypes.oneOf(constants.PLACEMENTS)
 };
 
-Caption.propDefault = {
+Caption.defaultProps = {
   placement: 'center'
 };
 
--- a/src/Slide.js
+++ b/src/Slide.js
@@ -1,31 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const Slide = ({ image, children }) => {
-  if (typeof image === 'string') {
-    return (
-      <li>
-        <img src={image} alt="" />
-        {children}
-      </li>
-    );
-  } else {
-    return (
-      <li>
-        {image}
-        {children}
-      </li>
-    );
-  }
-};
+const Slide = ({ image, children }) => (
+  <li>
+    {image}
+    {children}
+  </li>
+);
 
 Slide.propTypes = {
   className: PropTypes.string,
   children: PropTypes.node,
   /**
-   * The path of the background image
+   * The image that will be used in the Slide
    */
-  image: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired
+  image: PropTypes.node.isRequired
 };
 
 export default Slide;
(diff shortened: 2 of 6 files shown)


src/Caption.js
```diff
@@ -0,0 +1,38 @@
...
+
+Caption.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  /**
+   * Placement of the caption
+   * @default 'center'
+   */
+  placement: PropTypes.oneOf(constants.PLACEMENTS),
+  style: PropTypes.object
```
alextrastero: this `style` is not used?

src/Caption.js
```diff
@@ -0,0 +1,38 @@
...
+
+const Caption = ({ className, placement, children, ...props }) => {
+  return (
+    <div
+      className={cx(
+        'caption',
+        {
+          [`${placement}-align`]: placement
+        },
+        className
```
alextrastero: seems like placement is always defined, so this could be like:
```jsx
'caption',
`${placement}-align`,
className
```

src/Slide.js
```diff
@@ -1,64 +1,31 @@
...
-        {this.renderCaption()}
+        <img src={src} alt="" />
+        {children}
+      </li>
+    );
+  } else {
+    return (
+      <li>
+        {src}
+        {children}
```
alextrastero: I don't quite understand this, `<li>{src}{children}</li>`

alextrastero: why would `src` not be a string?

src/Caption.js
CloudPower97: Yeah, got merged with `{...props}`. An oversight of mine

src/Caption.js
CloudPower97: Nice catch

src/Slide.js
CloudPower97: Going to change this props from `src` to `image` and add an example to make it clear ASAP

CloudPower97: @alextrastero Done, please take a look at it whenever you wish

src/Slide.js
```diff
@@ -1,18 +1,18 @@
...
+        <img src={image} alt="" />
         {children}
       </li>
     );
   } else {
     return (
       <li>
-        {src}
+        {image}
         {children}
```
alextrastero: Sorry for this, but I still don't see how this could work, we would render an image next to the children? Can you point me to a materializecss example where this happens?

CloudPower97: Well, you will not find that in the documentation. 
You can have a look [here](https://github.com/Dogfalo/materialize/blob/v1-dev/js/slider.js#L58) to understand why this is going to work with a `picture` element.
Basically this allows one to use a [`picture`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) html element for [art direction](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction) or simply to provide different version of your images (WebP images, JPEG...)
For the full example please see my first comment.

CloudPower97: @alextrastero Let me know if you still want to use `src` as a prop name. I think `image` or `img` would be better. Or maybe having it both?
Like `src` is a `string` and it will handle the first case and `image` is a `node` and we are going to render it differently.
Basically it's the same thing is happening now under the hood, but doing so will not break existing code.

CloudPower97:
Hey @alextrastero, do you want me to do something here? Or can this be merged? :smiley:

test/Caption.spec.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Caption from '../src/Caption';
+
+describe('<Caption />', () => {
+  let wrapper = shallow(
+    <Caption placement="left">
```
alextrastero: what happens if we dont supply placement prop?

CloudPower97: It defaults to `center`

```javascript
Caption.defaultProps = {
  placement: 'center'
};
```

CloudPower97:
@alextrastero Here we go :+1:

PR #670, opened 2018-10-13
3.0.0-SideNav

CloudPower97:
# Description

This Pull Request is for #605 

Dropping 'jQuery' support, now using `M.Sidenav.init()`.

Adding new options (as per the [documentation](https://materializecss.com/sidenav.html#options)):

-  inDuration
- outDuration
- onOpenStart
- onOpenEnd
- onCloseStart
- onCloseEnd
- preventScrolling

Removing deprecated options:

- menuWidth
- closeOnClick

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

I've generated new snapshot and used the new mocker.

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have not generated a new package version. (the maintainers will handle that)

Issue #600, opened 2018-07-17
3.0.0 Navbar.js & NavItem.js


alextrastero:
### Upgrade to latest materialize-css@1.0.0-rc.2

AC
- Branch off 3.0.0-beta.0
- Check that the component works
- Update with new available modifiers and options
- Update propTypes and description
- Update examples page
- Create PR into branch 3.0.0-beta.0

PR #702, opened 2018-11-25
3.0.0-Navbar

CloudPower97:
# Description

This pull request is for #600 

Dropping _jQuery_ initialization, now using `M.Sidenav.init()`

For more information about the `Sidenav` please refer to #670.

Initializing the `Sidenav` component with the default options (as per[ the documentation](https://materializecss.com/sidenav.html#optons)):

Also dropping the use of `NavItem` component.
This was necessary to really make use of `React Router` `Link` component or any other custom component one may wish to use to fulfill routing purposes.

For the same reason, I'm dropping the `href` props, instead fully relying on `brand` props which in turns make it possible (again) to use `Link` component from `React Router` or whatever you are using for routing purposes.

For example:
```javascript
<Navbar brand='Logo' href='/' />
     ...
</Navbar>
```
Is now:

```javascript
<Navbar brand={<a href="/">Logo</a>} />
     ...
</Navbar>
```

With `Link` component from `React Router`:

```javascript
<Navbar brand={<Link to="/">Logo</Link>} />
     ...
</Navbar>
```
I think this approach is more flexible and resolve some edges cases, as in #650  

Moreover It's now possible to add extended components to the `NavBar`[ as per the doc](https://materializecss.com/navbar.html#navbar-tabs).

To completely use the `extended component` functionality I think we have to refactor, for example `Tabs` component, maybe using `Portals`

## Type of change

# How Has This Been Tested?

Used it in a project of mine and tested running `yarn test`.
I've updated the snapshot accordingly to the changes.

Also added some new tests too check 
- if the sidenav component initializes with default options if none are given
- if the mobile sidenav is rendered.
- it it's possible to center the logo using `centerLogo`
- can be extended with custom elements
# Checklist:

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -4,6 +4,5 @@ export default {
   SIZES: ['s', 'm', 'l', 'xl'],
   PLACEMENTS: ['left', 'center', 'right'],
   SCALES: ['big', 'small'],
-  ICON_SIZES: ['tiny', 'small', 'medium', 'large'],
-  MOBILE_NAV_ID: 'mobile-nav'
+  ICON_SIZES: ['tiny', 'small', 'medium', 'large']
 };
--- a/src/constants.js
+++ b/src/constants.js
@@ -4,5 +4,6 @@ export default {
   SIZES: ['s', 'm', 'l', 'xl'],
   PLACEMENTS: ['left', 'center', 'right'],
   SCALES: ['big', 'small'],
-  ICON_SIZES: ['tiny', 'small', 'medium', 'large']
+  ICON_SIZES: ['tiny', 'small', 'medium', 'large'],
+  MOBILE_NAV_ID: 'mobile-nav'
 };
--- a/test/Navbar.spec.js
+++ b/test/Navbar.spec.js
@@ -1,13 +1,26 @@
 import React from 'react';
 import { shallow, mount } from 'enzyme';
 import Navbar from '../src/Navbar';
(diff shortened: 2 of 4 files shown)


src/constants.js
```diff
@@ -4,5 +4,6 @@ export default {
   SIZES: ['s', 'm', 'l', 'xl'],
   PLACEMENTS: ['left', 'center', 'right'],
   SCALES: ['big', 'small'],
-  ICON_SIZES: ['tiny', 'small', 'medium', 'large']
+  ICON_SIZES: ['tiny', 'small', 'medium', 'large'],
+  MOBILE_NAV_ID: 'mobile-nav'
```
alextrastero: If this constant is not going to be used anywhere except `Navbar.js` maybe we could keep it there

test/Navbar.spec.js
```diff
@@ -1,47 +1,111 @@
...
   test('can be fixed', () => {
     wrapper = shallow(<Navbar fixed />);
     expect(wrapper).toMatchSnapshot();
-    expect(wrapper.hasClass('navbar-fixed')).toBeTruthy();
+    expect(wrapper.find('.navbar-fixed')).toHaveLength(1);
+  });
+
+  test('should have a sidenav component', () => {
+    wrapper = shallow(<Navbar />);
+    expect(wrapper.find('.sidenav')).toHaveLength(1);
```
alextrastero: This test is redundant because of test on L30

PR #711, opened 2018-12-13
clean up sidenav specs

alextrastero:
- no need for extra div, can use `Fragment`
- removed complex test setup method
- removed redundant tests


alextrastero:
can you have a look @CloudPower97 ?

src/SideNav.js
```diff
@@ -31,12 +31,12 @@ class SideNav extends Component {
     );
 
     return (
-      <div>
+      <Fragment>
```
CloudPower97: Since we are using `Babel 7` from #714 , we can also use the [`Fragment Short Syntax`](https://reactjs.org/docs/fragments.html#short-syntax)

Hence we can also drop the `Fragment` import at L1

CloudPower97:
@alextrastero Sure! I've added a comment :grinning:

PR #721, opened 2018-12-16
3.0.0-beta.0 Button Stories

onluiz:
# Description

Add Button.stories.js to 3.0.0-beta.0 version.
Issue #583

Variants:
- [x] Default
- [x] Floating
- [x] Flat
- [x] Submit
- [x] Large
- [x] Small
- [x] Disabled
- [x] Link
- [x] FAB Vertical
- [x] FAB Horizontal

## Type of change

# How Has This Been Tested?
- Using `npm run storybook` and checking results into the browser.

# Checklist:


stories/Button.stories.js
```diff
@@ -0,0 +1,81 @@
...
+  <Button floating fab='vertical' icon='edit' className='red' large>
+    <Button floating icon='insert_chart' className='red'/>
+    <Button floating icon='format_quote' className='yellow darken-1'/>
+    <Button floating icon='publish' className='green'/>
+    <Button floating icon='attach_file' className='blue'/>
+  </Button>
+);
+
+stories.add('Horizontal FAB', () => 
+  <Button floating fab='direction-left' icon='edit' className='green' large>
```
alextrastero:
```
warning.js:33 Warning: Failed prop type: Invalid prop `fab` of value `direction-left` supplied to `Button`, expected one of ["vertical","horizontal","toolbar"].
    in Button
```

stories/Button.stories.js
```diff
@@ -0,0 +1,81 @@
...
+  <div>
+    <Button large disabled style={{marginRight: '5px'}}>Button</Button>
+    <Button disabled style={{marginRight: '5px'}}>Button</Button>
+    <Button flat disabled style={{marginRight: '5px'}}>Button</Button>
+    <Button floating disabled><Icon>add</Icon></Button>
+  </div>
+);
+
+stories.add('Link', () => 
+  <Button waves='light' node='a' href='http://www.google.com'> Open Me In New Tab </Button>
```
alextrastero: let's link to `http://react-materialize.github.io/` and remove extra spaces

stories/Button.stories.js
```diff
@@ -0,0 +1,81 @@
...
+);
+
+stories.add('Fixed Action Button', () =>
```
alextrastero: This we could call `Vertical FAB`?

onluiz:
@alextrastero, it looks like there is even more variants to add to our Button component, like:
(https://materializecss.com/floating-action-button.html)
- FAB toolbar
- FAB direction to bottom
- FAB direction to right
- FAB click-only

Should I add those options too?

stories/Button.stories.js
alextrastero: This is what we want to mimic https://materializecss.com/floating-action-button.html#options

alextrastero:
Sure, let's add more examples :)

stories/Button.stories.js
onluiz: @alextrastero, added the options as it is on materializecss. Could you give me some feedback?

Here is the commit: https://github.com/react-materialize/react-materialize/pull/721/commits/fea436051242eaf8af802388866525b931092e90

Now we can use `top, bottom, left, right`. Vertical, horizontal and toolbar will work as usual.

Also, I have added the examples for these options.

src/Button.js
```diff
@@ -169,7 +186,11 @@ Button.propTypes = {
    * FAB Click-Only
    * Turns a FAB from a hover-toggle to a click-toggle
    */
-  fabClickOnly: PropTypes.bool
+  fabClickOnly: PropTypes.bool,
+  /**
+   * Styles
+   */
+  style: PropTypes.object
```
onluiz: For some reason styles weren't working (e.g `<Button fab="true" style={{...}}>...</Button>`). The problem only happens when using fab button, so I only adapted renderFab function.

Would it be nice if we create a `<FABButton>` or `<FixedFloatingButton>...<Button>...</Button><FixedFloatingButton>` in the future so we might have better organized component logic?

@alextrastero

alextrastero: The reason that `styles` or other dom attributes are not passed is because we specifically use `{...other}` on L79. So I would use a different approach: 

1. `renderFab` should read directly from `this.props` for that we might need to bind it in the constructor.
2. pass `{...other}` to the `div` like we do on L79

Bare in mind though that this PR should only handle one topic, and in this case it's adding the stories :) If we have an implementation bug let's tackle that separately with it's own specs/PR. Can we remove the commits for the `Button` changes and put them in another PR?

onluiz: I see.

Some of the stories (mainly fab ones) need those fixes in order to work. I will do as your advice and put those Button fixes into a different PR.

Thanks @alextrastero

alextrastero: I see, have a look here to see how you can add a style decorator to the stories that need it :)
https://github.com/react-materialize/react-materialize/blob/3.0.0-beta.0/stories/Collection.stories.js#L16

stories/Button.stories.js
```diff
@@ -60,35 +60,17 @@ stories.add('Link', () =>
   <Button waves='light' node='a' href='http://react-materialize.github.io/'>Open Me In New Tab</Button>
 );
 
-stories.add('Top FAB', () =>
-  <Button floating fab='top' icon='edit' className='red' large>
+stories.add('Vertical FAB', () =>
+  <Button floating fab='vertical' icon='edit' className='red' large>
```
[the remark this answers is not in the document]
onluiz: That PR will directly affect this one, since stories will change. We can close this one and I can create stories later or in that PR. What you think is the best approach?

:smile:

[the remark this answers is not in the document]
alextrastero: If this is a easy fix, I would keep the changes here. It would be easier to find in the future. Can this be done here?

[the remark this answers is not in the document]
alextrastero: Up to you @onluiz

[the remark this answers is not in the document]
onluiz: I fix stories in here when the other PR is merged then.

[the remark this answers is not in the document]
onluiz: @alextrastero, now I think we did it 😄 could you give a look at my last commit?

And thanks for your patience 😆

stories/Button.stories.js
```diff
@@ -0,0 +1,151 @@
...
+  </div>
+));
+
+stories.add('Link', () => (
+  <Button waves="light" node="a" href="http://react-materialize.github.io/">
+    Open Me In New Tab
+  </Button>
+));
+
+stories.add('Top FAB', () => (
```
alextrastero: let's categorize the fab stories for a neeter storybook, try to add `stories.add('FAB/Top'` Bottom, etc...

onluiz: Like this?

screen shot 2019-01-12 at 15 49 26

alextrastero:
Super!

PR #777, opened 2019-03-22
3.0.0 welcome story

alextrastero:
Adds basic description to `welcome.stories.js`, docs are quite ugly but as a first implementation it's ok. We can iterate as we go.


alextrastero:
@onluiz

onluiz:
Perfect :smile:

PR #800, opened 2019-04-01
fix: Datepicker now correctly handles onChange

CloudPower97:
# Description

While using `DatePicker` component I noticed a strange behavior.
When selecting a date, the `onChange` function seems to be called the next time the picker is drawn on the screen.
I've created a minimal, complete and verifiable example on CodeSandbox 

[Edit React Materialize DatePicker Bug](https://codesandbox.io/s/lmqp10wy7?fontsize=14)

This P.R fix this misbehavior. 

I also took the chance to remove unused code, improve `propTypes` and add `defaultProps`.

## Type of change

# How Has This Been Tested?

Added missing tests.
Added missing snapshot.

# Checklist:

### DIFF
--- a/src/DatePicker.js
+++ b/src/DatePicker.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import cx from 'classnames';
 import TextInput from './TextInput';
 import idgen from './idgen';
 
@@ -13,8 +12,13 @@ class DatePicker extends React.Component {
 
   componentDidMount() {
     if (typeof M !== 'undefined') {
+      const { onChange } = this.props;
       const elem = document.getElementById(this.id);
-      this.instance = M.Datepicker.init(elem, this.props.options);
+      const options = onChange
+        ? { onSelect: onChange, ...this.props.options }
+        : this.props.options;
+
+      this.instance = M.Datepicker.init(elem, options);
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/DatePicker.js
```diff
@@ -25,17 +29,9 @@ class DatePicker extends React.Component {
...
 
-    return (
-      <TextInput
-        id={this.id}
-        inputClassName="datepicker"
-        onSelect={onSelect}
-        {...other}
-      />
-    );
+    return <TextInput id={this.id} inputClassName="datepicker" {...other} />;
```
alextrastero:
```suggestion
    return <TextInput id={this.id} inputClassName="datepicker" {...this.props} />;
```

src/DatePicker.js
```diff
@@ -13,8 +12,13 @@ class DatePicker extends React.Component {
...
   componentDidMount() {
     if (typeof M !== 'undefined') {
+      const { onChange } = this.props;
       const elem = document.getElementById(this.id);
-      this.instance = M.Datepicker.init(elem, this.props.options);
+      const options = onChange
+        ? { ...this.props.options, onSelect: onChange }
+        : this.props.options;
+
+      this.instance = M.Datepicker.init(elem, options);
```
alextrastero: Nice! this could also be done like so, might be cleaner:

```jsx
const { onChange: () => {}, ...other } = this.props;
const elem = document.getElementById(this.id);

this.instance = M.Datepicker.init(elem, onSelect={onChange} {...other});
```

Is this cleaner or harder to understand?

src/DatePicker.js
```diff
@@ -106,7 +102,7 @@ DatePicker.propTypes = {
      * Number of years either side, or array of upper/lower range.
      * @default 10
      */
-    yearRange: PropTypes.oneOf([PropTypes.number, PropTypes.array]),
+    yearRange: PropTypes.oneOfType([PropTypes.number, PropTypes.array]),
```
alextrastero: 👍

CloudPower97:
> No need to change the Modal specs here :)

Not sure how this ended up in this commit :sweat_smile:

src/DatePicker.js
CloudPower97: I think it's a little bit harder to grasp, what do you think?

Issue #798, opened 2019-03-29
Assigning ClassNames Seems to Overwrite Materialize Classes


jamesmart77:
### Description

For the TextArea component, when we add classNames it only takes the custom classes we are providing and does not append them to the Materialize css...it simply overwrites Materialize. In v2.6.0 the React Materialize library would append our custom css, but the v3.0.0 is not behaving this way.

### Steps to Reproduce

1. <TextArea className: 'somethingCustom' />
2. Load component and you will see something similar to the screenshot provided.

**Expected behavior:** custom classNames should append to Materialize css 

**Actual behavior:** custom className is overwriting Materialize css

**Screenshots**
_Without Custom Css_
![image](https://user-images.githubusercontent.com/29242164/55258526-76210600-5239-11e9-8fe4-d50be6828e6a.png)

_With Custom Css_
![image](https://user-images.githubusercontent.com/29242164/55258546-83d68b80-5239-11e9-9c8a-0dc3fc0145b5.png)

### Versions

**react-materialize:** 3.0.3
**materialize-css:** 1.0.0
**react:** 16.8.2

onluiz:
@alextrastero, if it is a bug can I give a try?

alextrastero:
Thanks for reporting this @jamesmart77, yes please @onluiz :)

jamesmart77:
Thanks @onluiz for helping out with this! Saw this was included in a pull request and was wondering how this fix was going?

jamesmart77:
@onluiz @alextrastero any update on this? I can see release 3.0.4 #815 includes fix: Textarea now correctly handles layout classes. We just pulled down this update but are still seeing the same bug as before with the TextArea component.

**How we are using the Component**
![image](https://user-images.githubusercontent.com/29242164/56041936-ecc50580-5d07-11e9-93f5-1117b1a3c02e.png)

**UI Rendering**
![image](https://user-images.githubusercontent.com/29242164/56041991-09613d80-5d08-11e9-9a5d-057dbaa32a47.png)

alextrastero:
True, that wasn't merged yet, try `3.0.5`, it should work

PR #806, opened 2019-04-02
Fix textarea custom css classes

onluiz:
# Description

It was reported on issue #798 that Textarea component was losing its styles when custom classNames were passed to it. This pr fixes that behavior.

## Type of change

# How Has This Been Tested?

- unit testing
- browser testing

# Checklist:


src/Textarea.js
```diff
@@ -27,7 +35,7 @@ class Textarea extends Component {
           }}
           onChange={onChange}
           id={computedId}
-          className="materialize-textarea"
+          className={cx('materialize-textarea', className)}
           {...other}
```
alextrastero: one thing here @onluiz, could we first add `{...other}` and then overwrite with other props?
```jsx
{...other}
onChange={onChange}
id={computedId}
className={cx('materialize-textarea', className)}
```

onluiz: Done!

alextrastero:
Fixes #798

PR #814, opened 2019-04-04
Add missing Collapsible stories

CloudPower97:
# Description

As per the title

# Checklist:

### DIFF
--- a/stories/Collapsible.stories.js
+++ b/stories/Collapsible.stories.js
@@ -97,7 +97,7 @@ stories.add('Preselected Section', () => (
     <CollapsibleItem
       header="Yeah, you do seem to have a little 'shit creek' action going."
       icon="place"
-      className="active"
+      expanded
     >
       Yeah, you do seem to have a little 'shit creek' action going.
     </CollapsibleItem>
--- a/stories/Collapsible.stories.js
+++ b/stories/Collapsible.stories.js
@@ -14,7 +14,7 @@ stories.addParameters({
   }
 });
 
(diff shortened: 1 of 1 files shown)


stories/Collapsible.stories.js
```diff
@@ -61,3 +61,52 @@ stories.add('Expandable', () => (
...
+    <CollapsibleItem
+      header="Better safe than sorry. That's my motto."
+      icon="filter_drama"
+    >
+      Better safe than sorry. That's my motto.
+    </CollapsibleItem>
+    <CollapsibleItem
+      header="Yeah, you do seem to have a little 'shit creek' action going."
+      icon="place"
+      className="active"
```
alextrastero: should this be preselected? or how should we tackle this?

CloudPower97: @alextrastero I'm sorry, I don't grasp what are you trying to say :smile:

alextrastero: hehe sorry, shouldn't one of these components have an interface to **be** preselected?

CloudPower97: Sorry but I still don't understand :smile:

alextrastero: `CollapsibleItem` has a prop (interface) `expanded` which has the same effect of adding the class `active`, let's use that prop best :)

CloudPower97: Yep, you are right @alextrastero , sorry for the misunderstanding

Issue #811, opened 2019-04-04
Select does not put my data attribute into UL


uiktiomasfeliz:
### Description

The Select component does not work correctly with dynamic data.

![image](https://user-images.githubusercontent.com/5663564/55538689-4e112700-56bf-11e9-82e0-06fed856612c.png)

### Versions

**react-materialize:** 3.0.3
**materialize-css:** 1.0.0
**react:** 16.8.6

alextrastero:
Thanks for this, can you paste a code sample for me to test?

uiktiomasfeliz:
> Thanks for this, can you paste a code sample for me to test?


Here is an example similar to what I have in the project: [https://repl.it/repls/CostlyDishonestLanguage](https://repl.it/repls/CostlyDishonestLanguage)

CloudPower97:
@uiktiomasfeliz This should be fixed with `3.1.0`

alextrastero:
Closing, please let us know if it's not fixed :)

uiktiomasfeliz:
Hello, sorry for the delay, I updated to the latest version and the bug has not been solved, I leave the link of the demo: [https://repl.it/repls/PossibleRadiantTrace](https://repl.it/repls/PossibleRadiantTrace)

CloudPower97:
@uiktiomasfeliz You are right, #822 has not been merged as of now.
I'll work on it ASAP, hopefully it will be merged in the next release.
Thanks for the feedback. :blush:

PR #822, opened 2019-04-13
Select component with dynamic children

CloudPower97:
# Description

This __P.R__  is for #811 .

I took the occasion to also:

- `Select` now uses `Icon`
- Add `defaultProps`
- Make it possible to use `options` (as per the [docs](https://materializecss.com/select.html#options))
- Added a new `spec`
- Add a new `story` and reorganize the old ones

## Type of change

# How Has This Been Tested?

Added a `spec` to check if the `Select` component is initialized with default options if none are given.

# Checklist:

### DIFF
--- a/stories/Select.stories.js
+++ b/stories/Select.stories.js
@@ -13,8 +13,19 @@ stories.addParameters({
 });
 
 stories.add('Default', () => (
-  <Select onChange={action('Select changed')}>
-    <option value="" disabled selected>
+  <Select value="" onChange={action('Select changed')}>
+    <option value="" disabled>
+      Choose your option
+    </option>
+    <option value="1">Option 1</option>
+    <option value="2">Option 2</option>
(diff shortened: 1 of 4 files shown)


stories/Select.stories.js
```diff
@@ -13,8 +13,19 @@ stories.addParameters({
...
+      Choose your option
+    </option>
+    <option value="1">Option 1</option>
+    <option value="2">Option 2</option>
+    <option value="3">Option 3</option>
+  </Select>
+));
+
+stories.add('with Icon', () => (
+  <Select value="" icon="cloud">
```
alextrastero: can we remove `value=""`? if no value is passed it still works and like this it seems this is the required interface

CloudPower97: I used [materialize-css docs](https://materializecss.com/select.html) for reference.
Anyway, if you wish, I can remove it.

Issue #823, opened 2019-04-15
Select / Deselect Checkboxes Not Updating In React


jamesmart77:
### Select / Deselect Checkboxes Not Updating In React

We have a React component that lists checkboxes with user names. At the top of the list is a Select and Deselect All button. We have validated that our Redux store is updating and returning the appropriate props to the component so it should re-render the checkbox list either selected or deselected accordingly. The issue though is that each checkbox does not update, but remains its initial value.

### Steps to Reproduce

1. Create a React Application using Create-React-App
2. Map through an array of users and rendering the React-Materialize Checkbox component with the 'checked' parameter dynamic based on the users.isChecked value

**Expected behavior:** All checkboxes should be selected or deselected upon the click action of the corresponding button.

**Actual behavior:** Each checkbox does not update, but remains its initial value.

### Versions

**react-materialize:** 3.0.5
**materialize-css:** 1.0.0
**react:** 16.8.2

### Screenshot
![image](https://user-images.githubusercontent.com/29242164/56145780-ebe3dc00-5f72-11e9-8c3a-2fe8391423a2.png)

CloudPower97:
@jamesmart77 This should be fixed in version `3.1.0`

jamesmart77:
@CloudPower97 
Excited to try out this fix, but looks like there's an import directory reference error in src/Navbar.js

Please change line 5 in src/Navbar.js to `import TextInput from './TextInput';` I tried to submit a pull request but received a 403 on the branch push.

The problem seems to be that in the npm package the lib/Navbar.js file is pointing to a non-existent src directory. 

Line 16 in lib/Navbar.js of the npm v3.1.0 is where the problem is...
`var _TextInput = _interopRequireDefault(require("../src/TextInput"));`

CloudPower97:
@jamesmart77 Yep. Noticed this error earlier today; I already have a branch ready for that.
Thank you for reporting this.

Anyway, to avoid `403` have a look [here](https://github.com/react-materialize/react-materialize/blob/master/.github/CONTRIBUTING.md)

PR #832, opened 2019-04-17
Could not find module in path: 'react-materialize/src/TextInput' 

CloudPower97:
# Description

Since version `3.1.0`, as also noted by @jamesmart77 in #823, we have a `ModuleNotFoundError`:

```
Could not find module in path: 'react-materialize/src/TextInput' relative to '/node_modules/react-materialize/lib/Navbar.js'
```

This error has been unfortunately introduced in #813 

This __P.R__ fixes this

As a side note:
~How should we prevent this kind of error in the future, @alextrastero ?~
I've come up with this solution: 
using [`ESLint` `no-restricted-imports` rules
](https://eslint.org/docs/rules/no-restricted-imports) we can restrict the use of `../src/something`.

I've introduced this with [this `commit`](https://github.com/react-materialize/react-materialize/pull/832/commits/bfec950e325ded27c08419f70391bee74747f551).

Let me know if this is ok for you :smile:

## Type of change

# Checklist:

### DIFF
--- a/src/.eslintrc
+++ b/src/.eslintrc
@@ -1,0 +1,14 @@
+# JavaScript Standard Style Guide: http://standardjs.com
+# https://github.com/feross/eslint-config-standard-react
+rules:
+  no-restricted-imports: [
+    "error",
+    {
+      "paths": [
+        "../src/",
+      ],
+      "patterns": [
+        "../src/*",
+      ]
+    }
+  ]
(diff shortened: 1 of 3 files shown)


jamesmart77:
@alextrastero do you know when this fix will be merged? Because of previous bugs, this deployment past v3.1.0 is necessary for us to get off of v2.6.

CloudPower97:
@jamesmart77 Meanwhile you can clone the repo, make the fix locally and run `yarn link`; 
Then run `yarn link "react-materialize"` in the projects where you want to use this package.
You can also use `npm` but I've never done that with it.

alextrastero:
I'm not 100% with this change. We need to be a bit more strict and careful on reviews. Another solution would be to have a small smoke test with Puppeteer maybe.

PR #795, opened 2019-03-29
Allow the use of `picture` HTML element in Parallax component

CloudPower97:
# Description

As in #701 , you can now use a `picture` html element for art direction or simply to provide different version of your images (`WebP` images, `JPEG`...)

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

Add test to check if one can use a `picture` element
Updated every snapshots.

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have not generated a new package version. (the maintainers will handle that)

PR #834, opened 2019-04-18
It's now possible to use `picture` HTML element in Mediabox component

CloudPower97:
# Description

As in #795 and #701, it's now possible to use a `picture` `HTML` element or a component like [`react-progressive-picture`](https://github.com/CloudPower97/react-progressive-picture) or [`react-responsive-picture`](https://github.com/braposo/react-responsive-picture) and similar.

## Type of change

# How Has This Been Tested?

Added a new `spec`

# Checklist:

### DIFF
--- a/src/MediaBox.js
+++ b/src/MediaBox.js
@@ -1,13 +1,23 @@
-import React, { Component } from 'react';
+import { Component, cloneElement } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
+import idgen from './idgen';
 
 class MediaBox extends Component {
+  constructor(props) {
+    super(props);
+
+    this.id = props.id || `mediabox${idgen()}`;
+  }
+
   componentDidMount() {
     if (typeof M !== 'undefined') {
       const { options } = this.props;
 
-      this.instance = M.Materialbox.init(this._materialBoxed, options);
+      this.instance = M.Materialbox.init(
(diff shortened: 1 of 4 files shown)


src/MediaBox.js
```diff
@@ -1,13 +1,23 @@
...
+    this.id = props.id || `mediabox${idgen()}`;
+  }
+
   componentDidMount() {
     if (typeof M !== 'undefined') {
       const { options } = this.props;
 
-      this.instance = M.Materialbox.init(this._materialBoxed, options);
+      this.instance = M.Materialbox.init(
+        document.getElementById(this.id),
```
alextrastero: I know this change was also done in a previous PR but just for my understanding, why do we favour using an `id` instead of the actual node ref?

CloudPower97:
> What about having the image as a child?

Yeah, you are right, it's better to have it as a child.

Going to work on this ASAP.

CloudPower97:
@alextrastero Have a look at it whenever you want :smile:

CloudPower97:
@alextrastero ping :smile:

src/MediaBox.js
alextrastero: @CloudPower97, just for my understanding why is id better than ref node?

alextrastero: Ok, I guess it's to be able to use `cloneElement`

PR #861, opened 2019-05-06
Dynamically update `checked` prop for Switch component

### LINKED ISSUE #860: Switch does not handle attribute updates
### Description

The Switch component can be used perfectly, when it is clicked directly. The lever will toggle as expected. But in case that the switch is toggled via a different component (e.g. two switches share the same state) the "not-clicked" lever does not updated - although the checked attribute changes correctly in the DOM. 

### Steps to Reproduce

1. click on one of the Switches - only the clicked lever will update (visually)
```
constructor(props) {
  super(props);
  this.state = {
    cbk: false
  }
}
```

```	
render(){
  return(
    <div>
      <Switch
        checked={this.state.cbk}
        offLabel={"off"}
        onLabel={"on"}
        onChange={() => {
          this.setState({cbk: !this.state.cbk});
        }}
      />
      <Switch
        checked={this.state.cbk}
        offLabel={"off"}
        onLabel={"on"}
        onChange={() => {
          this.setState({cbk: !this.state.cbk});
        }}
      />
    </div>
  )
}
```

**Expected behavior:** 

As both switches refer to the same state both levers should be updated

**Actual behavior:** 

1) only the clicked lever updates automatically
2) both input fields will get the "checked" attribute

### Versions

**react-materialize:** 3.2.2
**materialize-css:** 1.0.0
**react:** 16.8.6

CloudPower97:
# Description

This __P.R__ fixes #860 

## Type of change

# How Has This Been Tested?

Tested with `Formik`

# Checklist:

### DIFF
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -60,7 +60,8 @@ Switch.propTypes = {
 };
 
 Switch.defaultProps = {
-  id: `switch-${idgen()}`
+  id: `switch-${idgen()}`,
+  onChange: () => {}
 };
 
 export default Switch;
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -1,9 +1,16 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import cx from 'classnames';
 import idgen from './idgen';
 
-const Switch = ({ checked, onChange, disabled, id, offLabel, onLabel }) => {
+const Switch = ({
+  id,
+  className,
+  checked,
+  onChange,
+  disabled,
+  offLabel,
+  onLabel
+}) => {
   const computedId = id || idgen();
 
   return (
@@ -12,10 +19,11 @@ const Switch = ({ checked, onChange, disabled, id, offLabel, onLabel }) => {
         {offLabel}
         <input
           id={computedId}
+          className={className}
           disabled={disabled}
           onChange={onChange}
           type="checkbox"
-          defaultChecked={checked}
+          checked={checked}
         />
         <span className="lever" />
         {onLabel}
@@ -25,6 +33,12 @@ const Switch = ({ checked, onChange, disabled, id, offLabel, onLabel }) => {
 };
 
 Switch.propTypes = {
+  /*
+   * override id
+   * @default idgen()
+   */
+  id: PropTypes.string,
+  className: PropTypes.string,
   /*
    * label for off
    */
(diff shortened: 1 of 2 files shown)


alextrastero:
Also, let's pass rest of props instead of classname ?

CloudPower97:
> Shouldn't we ALSO have `defaultChecked`? https://reactjs.org/docs/uncontrolled-components.html

Not sure about that... I mean, we have to choose between `controlled` or `uncontrolled` component.
I do not think we can implement both things.
But as I said before, I'm not completely sure about that.

Anyway, I've spreaded all `prop`s now.
I've also used `defaultProps` for `id` and removed `Component` import from `React` which is not used at all.
Have a look at this whenever you can :smile:

src/Switch.js
```diff
@@ -52,4 +59,8 @@ Switch.propTypes = {
   checked: PropTypes.bool
 };
 
+Switch.defaultProps = {
+  id: idgen()
```
alextrastero: 👍

src/Switch.js
```diff
@@ -1,21 +1,27 @@
...
+      <label htmlFor={id}>
         {offLabel}
         <input
-          id={computedId}
+          id={id}
           disabled={disabled}
           onChange={onChange}
           type="checkbox"
-          defaultChecked={checked}
+          checked={checked}
```
alextrastero: since we have spread, do we need `id, disabled, onChange, checked`

CloudPower97: Yes, we have to keep these `prop`s as is

CloudPower97:
> Regarding `defaultChecked` controlled vs uncontrolled I agree, but for `TextInput` we pass both, is this good?

According to [`React` docs ](https://reactjs.org/docs/uncontrolled-components.html)

> With an uncontrolled component, you often want React to specify the initial value, but leave subsequent updates uncontrolled.

So it seems that the `defaultChecked` `prop` is only used during initial render.

I'm still not sure if we can use both `defaultChecked` and `checked`.

To be honest, I'm inclined to the `controlled` component approach, so I think we should simply drop `defaultChecked`, also because one can achieve the same by setting `checked` to `true` or `false`, depending on your needs, mimicking `defaultChecked` or `defaultValue` in the case of `TextInput`.

What do you think we should do? :smiley:

CloudPower97:
@alextrastero ping :smile:

src/Switch.js
alextrastero: `idgen` returns a number, let's convert it to a string

alextrastero:
Yes agreed, I would prefer a controlled component, but then we'll need a default prop of `onChange: () => {}` to remove the warning.

alextrastero:
Maybe a follow up issue ticket to investigate if this can also be done in `<TextInput />`

alextrastero:
👍

PR #869, opened 2019-05-15
SideNavItem onClick

adamdriscoll:
# Description

I wanted to be able to hook up to the onClick handler of the SideNavItem rather than relying on an HREF. 

## Type of change

- [x ] New feature (non-breaking change which adds functionality)
- [x ] This change requires a documentation update

# How Has This Been Tested?

I added jest tests and tested the packaged component. 

# Checklist:

### DIFF
--- a/src/SideNavItem.js
+++ b/src/SideNavItem.js
@@ -14,6 +14,7 @@ class SideNavItem extends Component {
       waves,
       user = {},
       children,
+      onClick,
       ...props
     } = this.props;
     const itemClasses = {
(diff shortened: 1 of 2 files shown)


adamdriscoll:
Realized there was a UserView prop that would allow me to do this.

nckswt:
Is there a way to provide an `onClick` handler without having to go through UserView props? This would still be a useful feature. 

My own use case: an Electron app using React-Materialize without wanting to have to rely on things like React-Router.

alextrastero:
@nckswt I think this is a valid request, I reopened this and added some changes in #912

PR #875, opened 2019-06-05
Feature allow prop.id for Navbar

tm1000:
# Description

This adds the ability to define a custom ID for a navbar. This is useful in situations where-as the is a need for multiple navbars on the same page so that the mobile sidenav will work and not conflict because of duplicated ids

## Type of change

- [X] This change requires a documentation update

# How Has This Been Tested?

The default if no ID is defined for the Navbar component is to fallback to 'mobile-nav' thus there are no breaking changes

# Checklist:

### DIFF
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -98,6 +98,7 @@ class Navbar extends Component {
 }
 
 Navbar.propTypes = {
+  id: PropTypes.string,
   brand: PropTypes.node,
   children: PropTypes.node,
   className: PropTypes.string,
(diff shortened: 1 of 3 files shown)


tm1000:
@alextrastero doing it as switch does it, added a spec as well

tm1000:
This actually passed and Travis says so but the hook did not trigger back here.

alextrastero:
Let's fix the conflicts, the CI should rerun then and pass

tm1000:
Since merging #877 will break this as well I will wait to fix this when #877 is merged.

alextrastero:
Update this and i'll re-review looking good 💯

PR #877, opened 2019-06-05
Fix #876, clone sidenav elements and change their prop.id to append sidenav

### LINKED ISSUE #876: Using id prop in dropdown(s) inside of navbar causes dropdowns to not function
### Description

When utilizing Navar items with dropdowns and defining the dropdown ids through prop.id the dropdowns fail to render (well... they DO render but...). This only happens with custom ids, when you let the generator randomly generate ids there is no issue 

The issue is that the rendered dropdowns are cloned into the navbar and into the sidenav this causes id conflicts and so react doesnt know what to do and so the original dropdowns don't render

See: https://github.com/react-materialize/react-materialize/blob/master/src/Navbar.js#L65
and
https://github.com/react-materialize/react-materialize/blob/master/src/Navbar.js#L88

### Steps to Reproduce

```
<Navbar>
    <Dropdown id="this-is-my-id" options={{hover: true, coverTrigger: false, autoTrigger: false, constrainWidth: false}} trigger={<a>Test</a>}</Dropdown>
</Navbar>
```

Try to navigate to a dropdown 

**Expected behavior:** Dropdowns should render when the trigger is activated

**Actual behavior:** The dropdowns dont render (because the dropdown ids in the navbar and mobile navbar conflict)

### Versions

**react-materialize:** 3.3.1
**materialize-css:** 1.0.0
**react:** 16.8.6

tm1000:
# Description

If id is detected on child 'links' then use React.cloneElement and change the id to prefix sidenav onto it so it doesnt conflict with itself in the sidenav. See #876 for more information

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

I've tested this in my own environment to ensure the functionality still works as intended. It only affects children that have prop.id set.

# Checklist:


tm1000:
Add spec.

alextrastero:
Seems like you have pull requests that conflict with each other

tm1000:
@alextrastero yeah they all modify the same file so each time you merge them I'll have to go back and fix them :-) just GIT sometimes :-D

PR #878, opened 2019-06-05
Allows custom sidenav node for navbar

tm1000:
# Description

This request allows one to define a custom sidebar node when using a navbar. I ran into this issue when using dropdowns from the sidenav as they rendered really weird and ugly (not any fault of this package) so I thought it'd be nice to be able to define a custom node for the sidenav

## Type of change

- [X] This change requires a documentation update

# How Has This Been Tested?

This falls back to the old way if the sidenav prop is not defined

# Checklist:

### DIFF
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -45,6 +45,8 @@ class Navbar extends Component {
       <li key={index}>{link}</li>
     ));
 
+    const sidenavLinks = sidenav ? sidenav : links;
+
     let navbar = (
       <nav className={navCSS}>
         <div className="nav-wrapper">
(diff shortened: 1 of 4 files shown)


tm1000:
Hi @alextrastero can you explain what you mean by "spec"

Thanks

alextrastero:
spec means a **test specification** in short spec :)

src/Navbar.js
```diff
@@ -74,6 +75,8 @@ class Navbar extends Component {
       navbar = <div className="navbar-fixed">{navbar}</div>;
     }
 
+    const sidenavLinks = sidenav ? sidenav : links;
```
alextrastero: actually why not move this to L44?

tm1000: I moved it in the latest commit. It was here because I figured it should be next to the other logic statement (fixed) but makes no difference in the end

tm1000:
Updated PR added spec addressed concerns

Issue #882, opened 2019-06-13
Radio Button does not re-render when given a value


masonjwhite:
### Description

Hey folks. We are using a radio button in a custom survey application. The value for said radio button is being provided via a redux store.

As we update the redux store (once a user clicks a radio button), and the user moves onto the next question (with a new default value coming in from state), the radio button is not re-rendering to account for the new value.

Upon inspection of the element using the Chrome Developer tools I can indeed confirm that the "value" attribute on the element is mismatched with what the radio button is rendering.

### Steps to Reproduce

1. Create a radio button component with the value tied to a piece of state
2. Change the value in state (via onChange handler or whatever works)

**Expected behavior:** [What you expect to happen]
Radio button should re-render to match the updated value coming in from state.

**Actual behavior:** [What actually happens]
The radio button's value prop is updated but the component does not re-render.

### Versions

**react-materialize:** 3.1.1
**materialize-css:** 1.0.0
**react:** 16.8.2

masonjwhite:
Is there any update on this?

PR #883, opened 2019-07-01
Fix radio group checked prop

JimDaGuy:
# Description

Changed RadioGroup component. Replaced the `defaultChecked` property of inputs with the `checked` property. RadioGroup component was not properly re-rendering when sent a bad value such as `null` (further explained in testing section).

For some context on usage: 
https://www.w3schools.com/jsref/prop_checkbox_defaultchecked.asp
https://www.w3schools.com/tags/att_input_checked.asp

Should solve RadioGroup issues such as https://github.com/react-materialize/react-materialize/issues/882 . 

## Type of change

# How Has This Been Tested?

1. Set a RadioGroup's value to a valid value, such as 4 on a 1-5 scale.
2. Set the RadioGroup's value to any non-valid value, such as `null`.
3. RadioGroup will not update and it will not unselect the input. 

After changing the property passed to inputs from `defaultChecked` to `checked`, the inputs will only remain checked while the RadioGroup's value matches the input's value.

Ran `npm run test:unit -- -u` and then `npm run test`. All tests passing.

# Checklist:

### DIFF
--- a/src/RadioGroup.js
+++ b/src/RadioGroup.js
@@ -23,7 +23,7 @@ const RadioGroup = ({
             id={`radio${idx}`}
             value={radioItem.value}
             type="radio"
-            defaultChecked={radioItem.value === value}
+            checked={radioItem.value === value}
             name={name}
             onChange={onChange}
             disabled={disabled}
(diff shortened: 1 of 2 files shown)


alextrastero:
:D wrong button, thanks for the contribution!

Issue #866, opened 2019-05-08
Dropping Icon component


CloudPower97:
Hey, @alextrastero, I think we can make any icon `prop` a `node` and dont use `Icon` component ourselves in other components of `react-materialize`.
If the user wants, he can use the component shipped with `react-materialize`, otherwise he is able to use anythings he wants, like _Material Design Icons for React_.

This is the same approach used in #793 .

Let me know if this is ok for you, I'd like to open a P.R for this, if you agree. :smile:

alextrastero:
Nice, we could also limit it a bit, of course it's good that components can handle multiple situations and edge cases but I would vote to only allow `icon` prop to be a `node`.

CloudPower97:
Yep, It definitely need to be a `node` and nothing else.
Alright then, I'll be working on this ASAP :man_technologist:

alextrastero:
👍

PR #890, opened 2019-07-11
feat: allow use of different `Icon` component

CloudPower97:
# Description

This **P.R** is to address #866 

## Type of change

# How Has This Been Tested?

Added required `spec`s

# Checklist:


CloudPower97:
@alextrastero ping :)

src/Button.js
```diff
@@ -98,7 +98,7 @@ class Button extends Component {
           toolbar: fab && fab.toolbarEnabled
         })}
       >
-        <a className={classes}>{this.renderIcon()}</a>
+        <a className={classes}>{icon}</a>
```
alextrastero: what happens if `icon` is not supplied?

src/Pagination.js
```diff
@@ -111,12 +117,16 @@ Pagination.propTypes = {
   children: PropTypes.node,
   className: PropTypes.string,
   maxButtons: PropTypes.number,
-  onSelect: PropTypes.func
+  onSelect: PropTypes.func,
+  leftBtn: PropTypes.node.isRequired,
+  rightBtn: PropTypes.node.isRequired
```
alextrastero: are these two required?

src/Button.js
CloudPower97: It simply doesn't render the `icon`. The `a` nod will still be present in the `DOM` and the functionality are preserved.

src/Pagination.js
CloudPower97: Yes, they are; When the developer doesn't provide them as `prop` they must be there no matter what.
For this reason I've also added 

```javascript
{
  ...
  leftBtn: <Icon>chevron_left</Icon>,
  rightBtn: <Icon>chevron_right</Icon>,
}
```

to `Pagination.defaultProps`

alextrastero: if `defaultProps` is present that means they are not required AFAIK

src/Button.js
alextrastero: do we want that though? an empty `<a className="foo"></a>` tag??

CloudPower97: Without this empty tag the functionality is completely broken.
That's because there is a lot of code which depends on this `a` tag;
For example some `CSS` will not be applied nor the `backdrop` element (`<div class="fab-backdrop"></div>`) will be appended.
Have a look yourself [here](https://github.com/Dogfalo/materialize/blob/25b3ac858f458f8ec1d91f49b150138711c1000f/js/buttons.js#L38).

src/Pagination.js
CloudPower97: Well, if you think about it it make sense to use `isRequired` coupled with `defaultProps`.

This component require these two `prop`s no matter what, and if you do not provide them the `default` one are going to be used (which, by the way, are the icons used by `materialize-css`).

As you can see I'm using `defaultProps` as a way to enforce that these two `prop`s are there.

Hope I explained my self well :)

alextrastero: I still don't quite understand. AFAIK `isRequired` means that without the prop, the component will break. It is not required because even if I don't pass the prop, it will use a default prop.

src/Button.js
alextrastero: Can we add a snapshot test as to what happens when I don't supply an `icon`?

src/Pagination.js
CloudPower97: Alright then, going to remove `isRequired` :)

src/Button.js
CloudPower97: Fun fact, it seems that the `spec` we have now for the `FAB` don't use the `icon` `prop` already :grin:
Check it out [here](https://github.com/react-materialize/react-materialize/blob/b562c48061d8b163f5f3050c918f7dcc08fa6fe5/test/Button.spec.js#L159).

alextrastero: True! good catch, we might need to do some changes here. Either set a default icon prop or not render, we can't have a `a` without any children. Thoughts?

alextrastero: Yeah, I think the best solution is to add a `add` icon as a default prop

CloudPower97: Yes, I agree with you. Just one thing, I'm going to use `edit` instead of `add`, as shown on `materialize-css` [documentation](https://materializecss.com/floating-action-button.html).

alextrastero:
My bad, I didn't check https://github.com/react-materialize/react-materialize/pull/890/commits/e2b0d9fb0360f3e99fffd19842030b48191e787b, it was visible in the specs that button was now always having an icon.

Issue #897, opened 2019-08-05
Autocomplete doesn't work


Giuseppe98PG:
### Description

Autocomplete doesn't work

### Steps to Reproduce

1. Install React-Materialize
2. Import Autocomplete from 'react-materialize'
3. Copy and paste this example http://react-materialize.github.io/react-materialize/?path=/story/javascript-autocomplete--default

**Expected behavior:** The autocomplete now should work but it doesn't

**Actual behavior:** The SPA crashes or the autocomplete doesn't work without giving any feedback

### Versions

**react-materialize:** 3.3.4
**materialize-css:** 1.0.0
**react:** 16.8.6

alextrastero:
Are we sure that `materialize-css` javascript is being included? Does any other js component work or only Autocomplete fails?

prag-onlineshop:
Me too, having a problem in AutoComplete
error: Autocomplete expected '}' inside options
but i only copy what said here: 
http://react-materialize.github.io/react-materialize/?path=/story/javascript-autocomplete--default

React: 16.5.2
react-materialize: 3.3.4
materialize-css: 1.0.0

madnight:
I can confirm that problem. It think it should be labeled "bug" instead of "question".

alextrastero:
I have just tried with this sample and it works as expected.
`<Autocomplete options={{data: {'Gus Fring': null,'Saul Goodman': null,'Tuco Salamanca': 'https://placehold.it/250x250'}}} placeholder="Insert here" />`

Can you paste a code sample?

alextrastero:
![Screen Shot 2019-08-31 at 11 50 43 AM](https://user-images.githubusercontent.com/4152819/64062350-950b3600-cbe5-11e9-99e2-8acd35991f2b.png)

CloudPower97:
Hey @alextrastero, testing this I've found that when providing dynamic `data` the `Autocomplete` didn't update itself with the new `data` provided in the `options` `prop`.

PR #907, opened 2019-09-09
fix: call updateData when providing new data

CloudPower97:
# Description

Testing out #897 I've found that `Autocomplete` didn't update itself upon receiving new `data` using the `options` `prop`.

## Type of change

# How Has This Been Tested?

I've added a new `spec` ensuring that [`updateData` `method`](https://materializecss.com/autocomplete.html#methods) is called upon receiving new `data`.

# Checklist:

### DIFF
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -25,6 +25,16 @@ class Autocomplete extends Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.instance) {
+      const { data } = this.props.options;
+
+      if (prevProps.options.data !== data) {
+        this.instance.updateData(data);
+      }
+    }
+  }
+
   componentWillUnmount() {
     if (this.instance) {
       this.instance.destroy();
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -29,7 +29,7 @@ class Autocomplete extends Component {
     if (this.instance) {
       const { data } = this.props.options;
 
-      if (prevProps.options.data !== data) {
+      if (JSON.stringify(prevProps.options.data) !== JSON.stringify(data)) {
         this.instance.updateData(data);
       }
     }
--- a/test/Autocomplete.spec.js
+++ b/test/Autocomplete.spec.js
@@ -53,6 +53,7 @@ describe('<Autocomplete />', () => {
     beforeEach(() => {
       autocompleteInitMock.mockClear();
       autocompleteInstanceDestroyMock.mockClear();
+      autocompleteUpdateDataMock.mockClear();
     });
 
     afterAll(() => {
@@ -62,6 +63,7 @@ describe('<Autocomplete />', () => {
     test('calls Autocomplete', () => {
       mount(<Autocomplete />);
       expect(autocompleteInitMock).toHaveBeenCalledTimes(1);
+      expect(autocompleteUpdateDataMock).not.toHaveBeenCalled();
     });
 
     test('calls updateData when providing new data', () => {
--- a/test/Autocomplete.spec.js
+++ b/test/Autocomplete.spec.js
@@ -10,6 +10,7 @@ describe('<Autocomplete />', () => {
     Google: 'http://placehold.it/250x250',
     'Apple Inc': null
   };
+
   const componentId = 'testAutocompleteId';
 
   let wrapper;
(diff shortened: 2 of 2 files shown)


src/Autocomplete.js
```diff
@@ -25,6 +25,16 @@ class Autocomplete extends Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.instance) {
+      const { data } = this.props.options;
+
+      if (prevProps.options.data !== data) {
```
alextrastero:
```js
var a = { foo: 2, baz: 3 }
var b = { baz: 3, foo: 2 }
a === b // false
```

I think it's **ok** to `updateData` every time here

test/Autocomplete.spec.js
```diff
@@ -36,24 +37,38 @@ describe('<Autocomplete />', () => {
...
     };
     const restore = mocker('Autocomplete', autocompleteMock);
+
     beforeEach(() => {
       autocompleteInitMock.mockClear();
       autocompleteInstanceDestroyMock.mockClear();
     });
 
+    afterAll(() => {
+      restore();
```
alextrastero: let's move this in the `beforeEach` to make sure that the mocks are cleared for each test case

src/Autocomplete.js
CloudPower97: What if we use this method?

```
var a = { foo: 2, baz: 3 }
var b = { baz: 3, foo: 2 }

JSON.stringify(a) === JSON.stringify(b) //false (the order matters)
JSON.stringify(a) === JSON.stringify({ foo:2, baz: 3}) //true
```

Concerning performance it's OK if the object is relatively small.
Moreover this also works  for deep nested objects.

Considering all of this and that `data` object is at most of the form 
```
{
     "Apple": null,
     "Microsoft": null,
     "Google": 'https://placehold.it/250x250'
}
```
 it should be OK.

Let me know your thoughts :)

test/Autocomplete.spec.js
CloudPower97: Actually we clear the mocks in `beforeEach`. But I forgot to clear the `autocompleteUpdateDataMock`. Just added that.
Moreover I've added `expect(autocompleteUpdateDataMock).not.toHaveBeenCalled();` to the `calls Autocomplete` `test`.

src/Autocomplete.js
CloudPower97: I've also tested this approach with a `spec`. It seems to work just fine.
For now I'm going to keep it on my machine until you provide me with some feedback.
Or would you like me to push this modification? :relaxed:

alextrastero: I'm afraid we can't assume the order of an object, I would rather remove the checking

test/Autocomplete.spec.js
alextrastero: so do we need to keep the `afterAll` `restore`?

CloudPower97: Yep, have a look yourself at others `spec`. We always `restore()` `afterAll`

alextrastero: Cool, thanks for checking 👍

CloudPower97: ;)

src/Autocomplete.js
CloudPower97: We can't just always call `updateData`, otherwise the component is not going to properly work.
It would call `updateData` even when the user is typing.
I'd really be grateful if you could have a look at the code I've now

CloudPower97: ping @alextrastero

alextrastero: Sorry for the delay. We need to solve this in a way that if we supply the same object but with different order we don't want to call `updateData` correct?

alextrastero:
The main idea is that this test should pass:
```js
  const options1 = { data: { Google: 1, Facebook: 2 } };
  const options2 = { data: { Google: 1, Facebook: 2 } };

  const wrapper = shallow(<Autocomplete options={options1} />);

  expect(autocompleteInitMock).toHaveBeenCalledTimes(1);
  wrapper.setProps({ options: options2 });
  expect(autocompleteInitMock).not.toHaveBeenCalled();
```

CloudPower97:
@alextrastero Everything should be fine, now. Could you check it ? I've added a couple of `specs` as well :smile:

CloudPower97:
Proposed solution is to use this [external package](https://www.npmjs.com/package/fast-deep-equal).

CloudPower97:
@alextrastero ping :+1:

CloudPower97:
@alextrastero Can you take a look at it?

alextrastero:
Hey @CloudPower97, how about we use hooks, and add as a dependency `options.data` and that's it? I know we can't determine if the data has changed or not, but I think it's ok.

CloudPower97:
> Hey @CloudPower97, how about we use hooks, and add as a dependency `options.data` and that's it? I know we can't determine if the data has changed or not, but I think it's ok.

That's not the case.

If we have a `useEffect` with a dependency on `options`, it doesn't matter if only `options.data` has changed, it's going to trigger itself anyway.

I see two solution as of now:
- use this [external package](https://www.npmjs.com/package/fast-deep-equal) and deep compare the `option` `object` (Then using an approach [like this one](https://stackoverflow.com/questions/54095994/react-useeffect-comparing-objects#answer-54710710)).
- Expose a `data` `prop`. `options.data` is mapped to this `prop`. Have an `useEffect` on `data` that calls `updateData`.

The thing is, with the first approach we do not break the old code but we introduce an external dependency, on the other hand with the second approach we break the old code but we do not depend on an external package.

Let me know what would you like to do.

alextrastero:
But that's the thing, is it really that important that this method is not 100% optimized? can't we add options as a dependency and rerender whenever it changes?

okb1100:
Is this ever going to get merged?

alextrastero:
Solved in #1070

PR #910, opened 2019-09-18
Adding support for onChange event on the searchBar

BrightEntity:
# Description

It was not possible to catch the onChange event from the search TextInput of the navBar component.


alextrastero:

PR #914, opened 2019-09-27
Test spec and snapshot updated

GeoffMooreMedia:
for RadioGroup user defined ID

# Description

Please include a summary of the change and which issue is fixed/what was added. Please also include relevant motivation and context.

**Changes Made**
Updates the RadioGroup component to add an optional parameter, to the options, to provide a user-defined input id. If an id is not provided, it will default to the original `radio${idx}` id. 

**Motivation**
I have a reusable component utilizing the RadioGroup. I also have a comparison view that uses two instances of this component side-by-side. As written, RadioGroup was duplicating input ids. This caused selection changes made in the second window to only apply to the RadioGroup in the first window. This update allows me to specify unique ids for each instance. For example...
``{ label: 'RadioA', value: 'A', id: `RadioA_${this.props.windowID}` }``
## Type of change

# How Has This Been Tested?

Jest test spec and snapshots have been updated for RadioGroup to test for this new capability and all tests still pass. I have used this in my comparison component where I was having issues previously and confirmed it is now using unique ids and selection works correctly in both windows. 

**Test Code**
```
class ParentComponent extends React.Component {
    render() {
        return (
            <div className="parent">
                <ChildComponent></ChildComponent>
                <ChildComponent></ChildComponent>
            </div>
        )
    }
}

class ChildComponent extends React.Component {
    render() {
        return (
            <RadioGroup
                name={`group_${this.props.windowID}`}
                label="Radio Buttons"
                value={this.state.selected}
                onChange={this.handleButtonChange.bind(this)}
                options={[
                    { label: 'Option A', value: 'optA', id: `optA_${this.props.windowID}` },
                    { label: 'Option B', value: 'optB', id: `optB_${this.props.windowID}` },
                    { label: 'Option C', value: 'optC', id: `optC_${this.props.windowID}` }
                ]}
            />
        )
    }
}
```

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9618,14 +9618,14 @@
       "dev": true
     },
     "fsevents": {
-      "version": "1.2.4",
-      "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.4.tgz",
-      "integrity": "sha512-z8H8/diyk76B7q5wg+Ud0+CqzcAF3mBBI/bA5ne5zrRUUIvNkJY//D3BqyH571KuAC4Nr7Rw7CjWX4r0y9DvNg==",
+      "version": "1.2.9",
+      "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.9.tgz",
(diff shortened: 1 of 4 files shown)


alextrastero:

Issue #928, opened 2019-11-23
Use React 16.8 Hooks


alextrastero:
We would like to start using [hooks](https://reactjs.org/docs/hooks-intro.html) wherever possible instead of class instances.

It's a good opportunity for learning how to use hooks if you've never tried them out.

Note: Some tests will fail using `shallow`, if this is the case, try with `mount` for that specific test.

**HELP WANTED** 🙇‍♀

- [x] src/Range.js
- [x] src/SideNavItem.js
- [x] src/Chip.js
- [x] src/Collapsible.js
- [x] src/Section.js
- [x] src/Parallax.js
- [ ] src/TextInput.js
- [ ] src/Navbar.js
- [ ] src/Textarea.js
- [x] src/Preloader.js
- [x] src/SideNav.js
- [x] src/Icon.js
- [ ] src/Tabs.js
- [ ] src/Dropdown.js
- [ ] src/Autocomplete.js
- [ ] src/Select.js
- [x] src/UserView.js
- [ ] src/Pagination.js
- [ ] src/Table.js
- [ ] src/CardTitle.js
- [ ] src/Checkbox.js
- [x] src/MediaBox.js
- [ ] src/Button.js
- [ ] src/Carousel.js
- [ ] src/TimePicker.js
- [x] src/Slider.js
- [ ] src/Card.js
- [ ] src/Modal.js
- [x] src/Collection.js
- [x] src/Col.js
- [x] src/Footer.js
- [ ] src/DatePicker.js

alextrastero:
Sample implementation here: https://github.com/react-materialize/react-materialize/pull/930

alextrastero:
Please check https://github.com/react-materialize/react-materialize/pulls in case the component has already been taken, sorry for not reviewing faster, I wasen't expecting such a response 💪

alextrastero:
Merged some in https://github.com/react-materialize/react-materialize/pull/966
Thanks to everyone involved!

alextrastero:
Merged some in #967 
Thanks again 👏

alextrastero:
So, turns out there was an issue with conflicting react versions, that manifests as a hook error. Was eventually solved, but I had to bump up 5 patch versions :D... will continue with the merging of the open PR's ASAP

alextrastero:
Closing this issue in favour of individual issues

PR #930, opened 2019-11-23
Range: Convert to stateless and use hooks

alextrastero:
Part of #928

Use hooks instead of class instances.

Details below 👇

### DIFF
--- a/src/Range.js
+++ b/src/Range.js
@@ -1,31 +1,26 @@
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 
-class Range extends React.Component {
-  componentDidMount() {
-    if (typeof M !== 'undefined') {
-      this.instance = M.Range.init(this._range);
-    }
-  }
+const Range = props => {
+  const instance = useRef(null);
+  const rangeRef = useRef(null);
 
-  componentWillUnmount() {
-    this.instance && this.instance.destroy();
-  }
+  useEffect(() => {
+    instance.current = M.Range.init(rangeRef.current);
 
-  render() {
-    return (
-      <p className="range-field">
-        <input
-          type="range"
-          ref={el => {
-            this._range = el;
-          }}
-          {...this.props}
-        />
-      </p>
-    );
-  }
-}
+    return () => {
+      if (instance.current) {
+        instance.current.destroy();
+      }
+    };
+  }, [rangeRef.current, instance.current]);
+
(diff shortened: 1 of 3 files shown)


src/Range.js
```diff
@@ -1,31 +1,26 @@
...
 
-class Range extends React.Component {
-  componentDidMount() {
-    if (typeof M !== 'undefined') {
-      this.instance = M.Range.init(this._range);
-    }
-  }
+const Range = props => {
+  const instance = useRef(null);
+  const rangeRef = useRef(null);
```
alextrastero: instead of `this.instance` we use `useRef`, more info: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

src/Range.js
```diff
@@ -1,31 +1,26 @@
...
 
-  componentWillUnmount() {
-    this.instance && this.instance.destroy();
-  }
+  useEffect(() => {
+    instance.current = M.Range.init(rangeRef.current);
```
alextrastero: We assign `current` to the `Range` instance

src/Range.js
```diff
@@ -1,31 +1,26 @@
...
-        />
-      </p>
-    );
-  }
-}
+    return () => {
+      if (instance.current) {
+        instance.current.destroy();
+      }
+    };
```
alextrastero: If `useEffect` returns a function this will be called when unmounting, more info: https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup

src/Range.js
```diff
@@ -1,31 +1,26 @@
...
+  }, [rangeRef.current, instance.current]);
```
alextrastero: these are the dependencies of `useEffect`, it's the equivalent of checking if prevProps.foo are different then props.foo. This useEffect will only be called again if any dependency changes.

test/Range.spec.js
```diff
@@ -8,10 +8,10 @@ describe('<Range />', () => {
   const mockInit = jest.fn();
 
   const rangeMock = { init: mockInit };
-  const restore = mocker('Range', rangeMock);
+  mocker('Range', rangeMock);
 
   test('renders', () => {
-    wrapper = shallow(<Range min="10" max="99" />);
+    wrapper = mount(<Range min="10" max="99" />);
```
alextrastero: using `mount` instead of `shallow` since `shallow` won't trigger `hooks`

PR #931, opened 2019-11-23
DatePicker: Convert to stateless and use hooks

ttermechikov:
Part of #928

Use hooks instead of class instances.


src/DatePicker.js
```diff
@@ -1,41 +1,29 @@
...
+import React, { useState, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import TextInput from './TextInput';
 import idgen from './idgen';
 
-class DatePicker extends React.Component {
-  constructor(props) {
-    super(props);
+const DatePicker = props => {
+  const [id] = useState(props.id || `datepicker${idgen()}`);
```
alextrastero: How about using a memoized value for this?

```jsx
const id = useMemo(() => props.id || `datepicker${idgen()}`, [props.id]);
```

alextrastero: I would only use `useState` if it's a value that needs to be updated.

src/DatePicker.js
```diff
@@ -1,41 +1,29 @@
...
-
-      this.instance = M.Datepicker.init(elem, options);
-    }
-  }
+  useEffect(() => {
+    const { onChange } = props;
+    const options = onChange
+      ? { ...props.options, onSelect: onChange }
+      : props.options;
+    const elem = document.getElementById(id);
```
alextrastero: How about we use a ref instead of an id here? I know in the previous implementation this is what was there, but we can take the opportunity to improve this :)

test/DatePicker.spec.js
```diff
@@ -1,23 +1,23 @@
 import React from 'react';
-import { shallow, mount } from 'enzyme';
+import { mount } from 'enzyme';
```
alextrastero: If the test passes with `shallow` let's keep shallow, it's faster :)

src/DatePicker.js
alextrastero: https://github.com/react-materialize/react-materialize/pull/933 👀

ttermechikov: Do we need to change the 'TextInput' component so that its ref refers to parentsRef?
For example in TextInput:
  const textInputRef = props.parentRef || useRef(null);
and in the DatePicker:
  <TextInput ref={datePickerRef} />?

Or is there another solution?

alextrastero: Very good point, I missed that. Normally we would use `forwardRef` but as `TextInput` is not yet stateless let's keep the `id` for now.

test/DatePicker.spec.js
alextrastero: So we'll need to use both, shallow and mount. Mount to test that it calls XYZ on mount.

ttermechikov: I have updated this. Did you mean this? :)

src/DatePicker.js
```diff
@@ -1,41 +1,29 @@
...
+import React, { useMemo, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import TextInput from './TextInput';
 import idgen from './idgen';
 
-class DatePicker extends React.Component {
-  constructor(props) {
-    super(props);
+const DatePicker = props => {
+  const id = useMemo(() => props.id || `datepicker${idgen()}`, [props.id]);
```
alextrastero: You know what i've seen a lot lately? making props.id a default value of `datepicker${idgen()`, WDYT?

src/DatePicker.js
```diff
@@ -1,41 +1,29 @@
...
-      this.instance = M.Datepicker.init(elem, options);
-    }
-  }
+  useEffect(() => {
+    const { onChange } = props;
+    const options = onChange
+      ? { ...props.options, onSelect: onChange }
+      : props.options;
+    const elem = document.getElementById(id);
+    instance.current = M.Datepicker.init(elem, options);
```
alextrastero: Also something else i've seen :D no need to create a ref since the instance is only used inside this scope, so `const instance = M...` is enough

src/DatePicker.js
ttermechikov: Do you mean something like that:
"const DatePicker = ({ id:  `datepicker${idgen()}`, ...props}) => {" ?

ttermechikov: Do you mean something like that:
const DatePicker = ({ id = `datepicker${idgen()}`, ...props }) => {
?

src/DatePicker.js
```diff
@@ -1,41 +1,28 @@
...
-      const elem = document.getElementById(this.id);
-      const options = onChange
-        ? { ...this.props.options, onSelect: onChange }
-        : this.props.options;
-
-      this.instance = M.Datepicker.init(elem, options);
-    }
-  }
+  useEffect(() => {
+    const { onChange } = props;
```
alextrastero: let's deconstruct all required props in the component creation: `const Foo = ({ bar, baz }) =>`
and then we can be more specific in the hooks dependencies :)

src/DatePicker.js
```diff
@@ -1,41 +1,28 @@
...
+import React, { useMemo, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import TextInput from './TextInput';
 import idgen from './idgen';
 
-class DatePicker extends React.Component {
-  constructor(props) {
-    super(props);
+const DatePicker = props => {
+  const id = useMemo(() => props.id || `datepicker${idgen()}`, [props.id]);
```
alextrastero: sorry for the back and forth, but it's best to move id to default props
`id: datepicker${idgen()}` and this is done :)

PR #934, opened 2019-11-23
Section: convert to stateless

daubaris:
Converting `Section` component to stateless. #928

### DIFF
--- a/test/Section.spec.js
+++ b/test/Section.spec.js
@@ -1,17 +1,17 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { mount } from 'enzyme';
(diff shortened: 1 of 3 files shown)


test/Section.spec.js
```diff
@@ -1,17 +1,17 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { mount } from 'enzyme';
```
alextrastero: In this case, we don't need to use `mount` as we are not testing any methods that should run on mount, shallow will do, and it's faster :)

daubaris: Oh yes, that is true. I'll fix it real quick :).

PR #935, opened 2019-11-24
Parallax: convert to stateless and use hooks

CloudPower97:
# Description

Part of #928 

# How Has This Been Tested?

No need to change `specs` here, as the current one already uses `mount`.

# Checklist:

### DIFF
--- a/src/Parallax.js
+++ b/src/Parallax.js
@@ -11,7 +11,7 @@ const Parallax = ({ children, className, image, options, ...props }) => {
     return () => {
       instance && instance.destroy();
     };
-  }, [_parallax]);
+  }, [_parallax, options]);
 
   return (
     <div className={cx('parallax-container', className)} {...props}>
--- a/src/Parallax.js
+++ b/src/Parallax.js
@@ -6,18 +6,12 @@ const Parallax = ({ children, className, image, options, ...props }) => {
   const _parallax = useRef(null);
 
   useEffect(() => {
-    let instance = null;
-
-    if (_parallax.current) {
-      if (typeof M !== 'undefined') {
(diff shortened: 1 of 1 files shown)


src/Parallax.js
```diff
@@ -1,39 +1,33 @@
...
+const Parallax = ({ children, className, image, options, ...props }) => {
+  const _parallax = useRef(null);
+
+  useEffect(() => {
+    let instance = null;
+
+    if (_parallax.current) {
+      if (typeof M !== 'undefined') {
+        instance = M.Parallax.init(_parallax.current, options);
+      }
```
alextrastero: I know this was how it was before, but what do you think about removing this check and if for some reason `M` does not exist, it will show an error instead of silently failing?

src/Parallax.js
```diff
@@ -1,39 +1,33 @@
...
-class Parallax extends Component {
-  componentDidMount() {
-    if (typeof M !== 'undefined') {
-      const { options } = this.props;
-      this.instance = M.Parallax.init(this._parallax, options);
+const Parallax = ({ children, className, image, options, ...props }) => {
+  const _parallax = useRef(null);
+
+  useEffect(() => {
+    let instance = null;
```
alextrastero: true! this instance is not needed outside 👍

src/Parallax.js
CloudPower97: Okay, so you want to delete that check, right?

alextrastero: yeah :)

CloudPower97: Here we go :smile:

alextrastero:
`git merge master --no-ff`

PR #943, opened 2019-11-24
TimePicker: refactor to use hooks

CloudPower97:
# Description

Part of #928 

# How Has This Been Tested?

Updated `specs`.
As in #931 we are going to keep the `id` until we refactor `TextInput` to use `useRef` `hook`.

# Checklist:

### DIFF
--- a/src/TimePicker.js
+++ b/src/TimePicker.js
@@ -15,7 +15,7 @@ const TimePicker = ({ onChange, options, ...props }) => {
     return () => {
       instance && instance.destroy();
     };
-  }, [document.getElementById(id)]);
+  }, [id, onChange, options]);
 
   return <TextInput id={id} inputClassName="timepicker" {...props} />;
 };
--- a/src/TimePicker.js
+++ b/src/TimePicker.js
@@ -1,39 +1,24 @@
-import React from 'react';
+import React, { useEffect, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import TextInput from './TextInput';
 import idgen from './idgen';
 
-class TimePicker extends React.Component {
-  constructor(props) {
-    super(props);
+const TimePicker = ({ onChange, options, ...props }) => {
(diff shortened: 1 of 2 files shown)


alextrastero:
`git merge master --no-ff`

test/__snapshots__/TimePicker.spec.js.snap
```diff
@@ -4,57 +4,13 @@ exports[`<TimePicker /> accepts external \`id\` value 1`] = `
 <TextInput
   id="ID"
   inputClassName="timepicker"
-  options={
```
alextrastero: how come these values have disappeared from snapshots?

CloudPower97: Actually, this is a good sign. :stuck_out_tongue_closed_eyes: 
`options` should never be in `snapshots` otherwise it means that the rendered component on the `DOM` would be like

```html
<input class="timepicker" type="text" id="timepicker3" options="[object Object]" value="">
```

which is not correct.

As a matter of fact, here is how the component is now rendered:

```html
<input class="timepicker" type="text" id="timepicker4" value="">
```

CloudPower97: ping @alextrastero :smile:

src/TimePicker.js
```diff
@@ -1,39 +1,24 @@
...
+import React, { useEffect, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import TextInput from './TextInput';
 import idgen from './idgen';
 
-class TimePicker extends React.Component {
-  constructor(props) {
-    super(props);
+const TimePicker = ({ onChange, options, ...props }) => {
+  const id = useMemo(() => props.id, [props.id]);
```
alextrastero: let's move to default props as in other components

alextrastero:
Continued here #1068

PR #942, opened 2019-11-24
Carousel: refactor to function component and use hooks

CloudPower97:
# Description

Part of #928 

# How Has This Been Tested?

The only `spec` failing is this one:
```js
  describe('undefined M', () => {
    let __M;
    beforeEach(() => {
      __M = global.M;
      global.M = undefined;
    });
    afterEach(() => {
      global.M = __M;
    });
    test('doesnt throw without M', () => {
      expect(() =>
        mount(<Carousel options={{ fullWidth: true }} />)
      ).not.toThrow();
    });
  });
```

That's because, as also [requested here](https://github.com/react-materialize/react-materialize/pull/935#discussion_r349933957), we have removed the check for `typeof M !== 'undefined'.

Let me know if you want to remove the `spec` or reintroduce the check.

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have not generated a new package version. (the maintainers will handle that)

PR #945, opened 2019-11-24
Textarea: use hooks

CloudPower97:
# Description

Part of #928 

# How Has This Been Tested?

As in #942 , let me know if you want to drop this specific `spec`
```js
  describe('undefined M', () => {
    let __M;
    beforeEach(() => {
      __M = global.M;
      global.M = undefined;
    });
    afterEach(() => {
      global.M = __M;
    });
    test('doesnt throw without M', () => {
      const element = mount(<Textarea />);
      expect(() => element.setProps({ label: 'foo' })).not.toThrow();
    });
  });
```

or reintroduce the check for `undefined M`.

# Checklist:

### DIFF
--- a/src/Textarea.js
+++ b/src/Textarea.js
@@ -1,123 +1,111 @@
-import React, { Component } from 'react';
+import React, { cloneElement, useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import idgen from './idgen';
 import constants from './constants';
 
-class Textarea extends Component {
-  constructor(props) {
-    super(props);
+const Textarea = ({
+  id,
+  children,
+  className,
+  s,
+  m,
+  l,
+  xl,
+  disabled,
+  noLayout,
+  placeholder,
+  icon,
+  label,
+  success,
+  error,
+  iconClassName,
+  onChange,
+  validate,
+  defaultValue,
+  value,
+  ...props
+}) => {
+  const _textarea = useRef(null);
+  const sizes = { s, m, l, xl };
 
-    this.id = props.id || idgen();
+  let responsiveClasses;
+  if (!noLayout) {
+    responsiveClasses = { col: true };
+    constants.SIZES.forEach(size => {
+      responsiveClasses[size + sizes[size]] = sizes[size];
+    });
   }
 
-  componentDidMount() {
-    if (typeof M !== undefined) {
-      // eslint-disable-next-line react/prop-types
-      this.props['data-length'] && M.CharacterCounter.init(this._textarea);
+  const wrapperClasses = cx('input-field', responsiveClasses);
+
+  const textareaProps = {
+    placeholder,
+    id,
+    value,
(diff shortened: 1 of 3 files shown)


alextrastero:
`git merge master --no-ff`

alextrastero:
The thing about the check. If M is not there, we want it to break :)

src/Textarea.js
```diff
@@ -1,123 +1,111 @@
...
+    id,
+    value,
+    defaultValue,
+    disabled,
+    ...props
+  };
+
+  useEffect(() => {
+    // eslint-disable-next-line react/prop-types
+    if (props['data-length']) {
```
alextrastero: Now that i'm seeing this, what is `data-length`? :D

src/Textarea.js
```diff
@@ -1,123 +1,111 @@
...
-      constants.SIZES.forEach(size => {
-        responsiveClasses[size + sizes[size]] = sizes[size];
-      });
+  useEffect(() => {
+    if (_textarea.current) {
+      M.textareaAutoResize(_textarea.current);
     }
+  }, [value]);
+
+  const renderLabel = () =>
```
alextrastero: useMemo? we only want to recompute this if label has changed
`const renderLabel = useMemo(() => ...`

alextrastero:
Thx so much @CloudPower97 for all the support :)

src/Textarea.js
CloudPower97: `data-length` is used to correctly initializes [`M.CharacterCounter`](https://materializecss.com/text-inputs.html#character-counter).

From [React Documentation](https://reactjs.org/docs/dom-elements.html) :

> The exception is aria-* and data-* attributes, which should be lowercased. For example, you can keep aria-label as aria-label.

Hence, as dashes are not allowed in variable names, we can't use `props.data-length` but we have to use quotes.

src/Textarea.js
CloudPower97: Thing is, `renderLabel` depends on `[error, id, label, placeholder, success, value]`.
Moreover it might be better to [`useCallback`](https://it.reactjs.org/docs/hooks-reference.html#usecallback) instead of `useMemo`.

But anyway, [I'm not really sure if we should use them](https://kentcdodds.com/blog/usememo-and-usecallback).

If we really want to do that I suggest to implement these changes in a next release, prior benchmarking  them to see the performances gain, if any.

Let me know your thoughs :wink:

src/Textarea.js
alextrastero: AHA! thanks for the clarification 👍

src/Textarea.js
alextrastero: Ah, ok, i've just read this. And i've been commenting in all the PR's to add `useMemo`. Ok let me have a thought and i'll get back to you tomorrow.

src/Textarea.js
CloudPower97: No problem :wink:

PR #950, opened 2019-11-25
Merge pull request #950 from CloudPower97/refactor/checkbox

CloudPower97:
# Description

Part of #928

Also fixed a bug which prevented the `checkbox` in a `checked` state if initially `checked`.
To do so, I used `useState` to track changes of `checked`.

## Type of change

# How Has This Been Tested?

Just a little note here, I moved the `idgen` call to a `defaultProps`.
For some reason it seems that this is failing:
```js
    expect(idgen).toHaveBeenCalledTimes(1);
```

But having a look at the `snapshots` everything looks just fine.
I decided to the test the effect of calling `idgen` instead of checking if `idgen` was called doing so:
```js
    expect(wrapper.prop('id')).toBe(`checkbox_${idgen()}`);
```

# Checklist:

### DIFF
--- a/src/Checkbox.js
+++ b/src/Checkbox.js
@@ -1,69 +1,52 @@
-import React, { Component } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import idgen from './idgen';
 
-class Checkbox extends Component {
-  constructor(props) {
-    super(props);
-  }
+const Checkbox = ({
+  id,
+  className,
+  indeterminate,
+  filledIn,
+  label,
+  onChange,
+  ...props
+}) => {
+  const [checked, setChecked] = useState(props.checked);
+  const _input = useRef(null);
 
-  componentDidMount() {
-    const { indeterminate } = this.props;
-
-    if (this._input && indeterminate) {
-      this._input.indeterminate = indeterminate;
-      this._input.checked = false;
+  useEffect(() => {
+    if (_input.current && indeterminate) {
+      _input.current.indeterminate = indeterminate;
+      _input.current.checked = false;
     }
-  }
+  }, [indeterminate]);
 
-  componentDidUpdate(prevProps) {
-    if (prevProps.indeterminate !== this.props.indeterminate) {
-      this._input.indeterminate = this.props.indeterminate;
-    }
-  }
+  delete props.checked;
 
-  render() {
-    const {
-      id,
-      className,
-      indeterminate,
-      filledIn,
-      disabled,
-      onChange,
-      checked,
-      value,
-      label
-    } = this.props;
-
-    let computedId = id || idgen();
-
-    if (indeterminate) {
-      computedId = `indeterminate-checkbox-${idgen()}`;
-    }
-
-    return (
-      <label htmlFor={computedId}>
-        <input
-          id={computedId}
-          className={cx(
-            {
-              'filled-in': filledIn
-            },
-            className
-          )}
-          disabled={disabled}
-          onChange={onChange}
-          type="checkbox"
-          checked={checked}
-          value={value}
-          ref={input => (this._input = input)}
-        />
-        <span>{label}</span>
-      </label>
-    );
-  }
-}
+  return (
+    <label htmlFor={id}>
+      <input
+        id={id}
+        className={cx(
+          {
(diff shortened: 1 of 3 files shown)


src/Checkbox.js
```diff
@@ -1,69 +1,52 @@
...
+const Checkbox = ({
+  id,
+  className,
+  indeterminate,
+  filledIn,
+  label,
+  onChange,
+  ...props
+}) => {
+  const [checked, setChecked] = useState(props.checked);
```
alextrastero: let's deconstruct checked, and that way we dont need to delete on L25

alextrastero: also, what happens if we pass checked as a new prop to this component, will it update?

CloudPower97: > let's deconstruct checked, and that way we dont need to delete on L25

I've done this by purpose. Otherwise we have to change the `checked` of `useState` to another, less semantic, name.

What do you suggest?

alextrastero: defaultChecked?

alextrastero:
About the calling of idgen https://stackoverflow.com/questions/52448482/how-to-test-default-props-functions-is-attached-to-component-or-not-react ?

src/Checkbox.js
CloudPower97: Well, I thought about `defaultChecked`, the thing is that it's misleading. 
We use it to track the `current` state of the `checkbox`, not the `default` one, as this name might suggest. :thinking: 
What do you think?

CloudPower97: > also, what happens if we pass checked as a new prop to this component, will it update?

Good point, I missed this one! :man_facepalming: 
It seems we have to `useEffect` here! :smile:

src/Checkbox.js
```diff
@@ -1,69 +1,55 @@
...
-        />
-        <span>{label}</span>
-      </label>
-    );
-  }
-}
+  return (
+    <label htmlFor={id}>
+      <input
+        {...props}
```
CloudPower97: @alextrastero Moving `{...props}` here allow us to avoid `delete props.checked`

src/Checkbox.js
```diff
@@ -1,69 +1,55 @@
...
-      className,
-      indeterminate,
-      filledIn,
-      disabled,
-      onChange,
-      checked,
-      value,
-      label
-    } = this.props;
+  useEffect(() => {
```
CloudPower97: @alextrastero We now update the state whenever `props.checked` changes

src/Checkbox.js
```diff
@@ -1,69 +1,55 @@
...
-  }
-
-  componentDidUpdate(prevProps) {
-    if (prevProps.indeterminate !== this.props.indeterminate) {
-      this._input.indeterminate = this.props.indeterminate;
+  useEffect(() => {
+    if (_input.current) {
+      _input.current.indeterminate = indeterminate;
+      _input.current.checked = false;
+      setChecked(false);
```
CloudPower97: @alextrastero We now correctly set the `state` to `false` when changing `undeterminate`

CloudPower97:
> About the calling of idgen https://stackoverflow.com/questions/52448482/how-to-test-default-props-functions-is-attached-to-component-or-not-react ?

Thank you for the suggestion.
Anyway, I was not trying to test if `idgen` is attached as a `defaultProp` (it isn't after all).

Instead, I'm not sure why `Jest` says that the function was not called when it obviously is, as clearly visible in both the `snapshot`s and the following `assert` that we do:

```js
    expect(wrapper.prop('id')).toBe(`checkbox_${idgen()}`);
```

I suppose it's a `Jest` quirks or perhaps the `defaultProp` is out of its scope, not really sure.

test/Checkbox.spec.js
```diff
@@ -13,8 +13,8 @@ describe('<Checkbox />', () => {
   });
 
   test('renders', () => {
-    wrapper = shallow(<Checkbox value="red" label="red" />);
-    expect(idgen).toHaveBeenCalledTimes(1);
+    wrapper = mount(<Checkbox value="red" label="red" />);
+    expect(wrapper.prop('id')).toBe(`checkbox_${idgen()}`);
     expect(wrapper).toMatchSnapshot();
```
alextrastero: let's only keep the snapshot and remove any check to idgen()

PR #953, opened 2019-11-26
Autocomplete: use hooks

CloudPower97:
# Description

Part of #928 

We might want to kill two birds with one stone and close #907 as well.

I left a comment there, let me know what do you want to do.

## Type of change

As `react-materialize` is a wrapper around `materialize` one should not rely on `onChange` when clicking on an `Autocomplete` suggestion.
For this, you should use `onAutocomplete`  `callback` that can be defined in `options`.
Moreover, having `onChange` returning the value instead of the `Synthetic Event` was really strange and not a thing that one expected (not me at least).

For these reasons, this is a breaking change.

# How Has This Been Tested?

Update `snapshot`

# Checklist:

### DIFF
--- a/src/Autocomplete.js
+++ b/src/Autocomplete.js
@@ -1,111 +1,66 @@
-/* eslint no-unused-vars: ["error", { "ignoreRestSiblings": true }] */
-import React, { Component } from 'react';
+import React, { cloneElement, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import constants from './constants';
 import idgen from './idgen';
 
-class Autocomplete extends Component {
-  constructor(props) {
-    super(props);
+const Autocomplete = ({
+  id,
+  className,
+  title,
+  icon,
+  s,
+  m,
+  l,
+  xl,
+  offset,
+  placeholder,
+  // these are mentioned here only to prevent from getting into ...props
+  value,
+  onChange,
+  options,
+  ...props
+}) => {
+  const _autocomplete = useRef(null);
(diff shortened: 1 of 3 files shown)


alextrastero:
That way we wouldn't be adding more breaking changes

CloudPower97:
> About `onChange` could we do something like this for the time being:
> 
> ```js-jsx
> const handleFoo = (e) => {
>   options.onAutocomplete(e);
>   
>   if (props.onChange) {
>     props.onChange(e);
>   }
> }
> const fooOptions = { ...options, onAutocomplete: handleFoo }
> ```

If we do that, `onChange` is going to receive the `value` instead of the `Synthetic Event`.
That's the whole point for this breaking change (and also to be closer to plain `materialize` usage).
I get your point but I really think that this change needs to be done.
Let me know :smile:

alextrastero:
But the `onAutomcomplete` is receiving whatever materialize is returning, so it is closer to materialize usage right? We're not tampering with the values, just passing them on, correct me if i'm wrong :)

CloudPower97:
> But the `onAutomcomplete` is receiving whatever materialize is returning, so it is closer to materialize usage right? We're not tampering with the values, just passing them on, correct me if i'm wrong :)

You might be right.
I'm going to double-check a couple of things.
I'll be back to you ASAP

alextrastero:
👋

alextrastero:
Hi @CloudPower97 any ideas if we can push forward this PR maybe together with #907 ?

alextrastero:
Continued here: #1070

PR #954, opened 2019-11-26
Collection: convert to stateless

daubaris:
# Checklist:
#928 


test/Collection.spec.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
-import { shallow, mount } from 'enzyme';
+import { mount } from 'enzyme';
```
alextrastero: did we confirm that these specs fail with shallow? :)

daubaris: Yeah, it's fine with shallow, restored it.

PR #955, opened 2019-11-26
Card: convert to stateless

daubaris:
# Checklist:
#928 

### DIFF
--- a/src/Card.js
+++ b/src/Card.js
@@ -1,20 +1,21 @@
-import React, { Component, Fragment, cloneElement } from 'react';
+import React, { Fragment, cloneElement } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import Icon from './Icon';
 
-class Card extends Component {
-  constructor(props) {
-    super(props);
-    this.renderTitle = this.renderTitle.bind(this);
-    this.renderReveal = this.renderReveal.bind(this);
-    this.renderAction = this.renderAction.bind(this);
-    this.renderContent = this.renderContent.bind(this);
-    this.renderAll = this.renderAll.bind(this);
-  }
-
-  renderTitle(title, reveal) {
-    const { revealIcon } = this.props;
+const Card = ({
+  title,
+  header,
+  className,
+  textClassName,
+  actions,
+  reveal,
+  children,
+  horizontal,
+  ...other
+}) => {
+  const renderTitle = (title, reveal) => {
+    const { revealIcon } = other;
 
     return (
       <span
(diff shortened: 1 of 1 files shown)


src/Card.js
```diff
@@ -1,20 +1,21 @@
...
+  header,
+  className,
+  textClassName,
+  actions,
+  reveal,
+  children,
+  horizontal,
+  ...other
+}) => {
+  const renderTitle = (title, reveal) => {
```
alextrastero: `const renderTitle = useMemo(...`

src/Card.js
```diff
@@ -26,10 +27,10 @@ class Card extends Component {
         {reveal && { revealIcon }}
       </span>
     );
-  }
+  };
 
-  renderReveal(title, reveal) {
-    const { closeIcon } = this.props;
+  const renderReveal = (title, reveal) => {
```
alextrastero: useMemo

src/Card.js
```diff
@@ -1,20 +1,21 @@
...
+  className,
+  textClassName,
+  actions,
+  reveal,
+  children,
+  horizontal,
+  ...other
+}) => {
+  const renderTitle = (title, reveal) => {
+    const { revealIcon } = other;
```
alextrastero: why not deconstruct?

src/Card.js
```diff
@@ -26,10 +27,10 @@ class Card extends Component {
         {reveal && { revealIcon }}
       </span>
     );
-  }
+  };
 
-  renderReveal(title, reveal) {
-    const { closeIcon } = this.props;
+  const renderReveal = (title, reveal) => {
+    const { closeIcon } = other;
```
alextrastero: why not deconstruct?

src/Card.js
```diff
@@ -40,63 +41,49 @@ class Card extends Component {
...
+  };
 
-  renderAction(actions) {
+  const renderAction = actions => {
     return <div className="card-action">{actions}</div>;
-  }
+  };
 
-  renderContent(title, reveal, textClassName, children) {
+  const renderContent = (title, reveal, textClassName, children) => {
```
alextrastero: actually I wonder, could we extract these methods outside of the component definition? since they're all getting values from args

daubaris: But it's using methods that are inside the component.

alextrastero: True, i'm just wondering if all these methods receive all the required props as arguments, why not move them outside the component render? .i.e

```jsx
const renderBar = (title) => {
  <p>{title}</p>
}

const Foo = (props) => {
  return (
    <div>
      {renderBar(props.title)}
    </div>
  )
}
```

WDYT?

daubaris: Well, if the function would be reused somewhere else in the project, then yeah, I'd see a point in moving it outside of a component render. But since they belong only to this component I don't really see the advantages, but then again maybe I am missing some details.

alextrastero:
Continued here #1064

PR #956, opened 2019-11-26
Dropdown: use hooks

CloudPower97:
# Description

Part of #928

# How Has This Been Tested?

The only test failing at this moment is the following one:

```js
  test('does not update ID reference with each render', () => {
    wrapper = shallow(<Dropdown trigger={<span>hi</span>} />);
    const { id } = wrapper.instance();
    wrapper.render();
    expect(wrapper.instance().id).toEqual(id);
  });
```

Citing [Enzyme documentation](https://airbnb.io/enzyme/docs/api/ShallowWrapper/instance.html#instance--reactcomponent):

> NOTE: can only be called on a wrapper instance that is also the root instance. With React 16 and above, instance() returns null for stateless functional components.

Anyway, I do not think this test is useful anymore, as `idx` has been removed.

# Checklist:

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -1,74 +1,56 @@
-import React, { Component, Fragment, Children, cloneElement } from 'react';
+import React, {
+  Fragment,
+  Children,
+  cloneElement,
+  useRef,
+  useEffect
+} from 'react';
 import PropTypes from 'prop-types';
 import idgen from './idgen';
 import cx from 'classnames';
 
-class Dropdown extends Component {
-  constructor(props) {
-    super(props);
-    this.idx = props.id || `dropdown${idgen()}`;
-    this.renderTrigger = this.renderTrigger.bind(this);
-    this.renderItems = this.renderItems.bind(this);
-  }
+const Dropdown = ({ children, className, trigger, options, ...props }) => {
+  const _dropdownContent = useRef(null);
 
-  componentDidMount() {
-    const { options } = this.props;
-
-    if (typeof M !== undefined) {
-      this.instance = M.Dropdown.init(
-        document.querySelectorAll(this._trigger),
-        options
-      )[0];
-    }
-  }
-
-  componentWillUnmount() {
-    if (this.instance) {
-      this.instance.destroy();
-    }
-  }
-
-  render() {
-    const { className, ...props } = this.props;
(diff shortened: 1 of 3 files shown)


src/Dropdown.js
```diff
@@ -1,74 +1,56 @@
...
-    const { trigger } = this.props;
+    return () => {
+      instance && instance.destroy();
+    };
+  }, [options, props.id]);
 
-    return cloneElement(trigger, {
-      'data-target': this.idx,
-      ref: t => (this._trigger = `[data-target=${this.idx}]`),
+  const renderTrigger = () =>
```
alextrastero: useMemo wherever possible

test/DropdownButton.spec.js
```diff
@@ -12,12 +12,12 @@ describe('<Dropdown />', () => {
...
-    const { idx } = wrapper.instance();
-    wrapper.render();
-    expect(wrapper.instance().idx).toEqual(idx);
-  });
+  // test('does not update ID reference with each render', () => {
+  //   wrapper = shallow(<Dropdown trigger={<span>hi</span>} />);
+  //   const { id } = wrapper.instance();
+  //   wrapper.render();
+  //   expect(wrapper.instance().id).toEqual(id);
+  // });
```
alextrastero: no need for this test as we've handled this by moving idgen to default props

CloudPower97: Yep, you are right! :smiley:

src/Dropdown.js
```diff
@@ -1,74 +1,56 @@
...
+      'data-target': props.id,
       className: cx(trigger.props.className, 'dropdown-trigger')
     });
-  }
-
-  renderItems() {
-    const { children } = this.props;
 
-    return Children.map(children, element => {
+  const renderItems = () =>
```
alextrastero: I think using a memoized helper here is a gain. Since it won't need to map through all children each render :)

CloudPower97: Out of curiosity, why do you want to `useMemo` and not `useCallback`?

CloudPower97: ping @alextrastero :smile:

alextrastero: `useMemo` :)

alextrastero:
Continued here #1066

PR #957, opened 2019-11-26
Select: use hooks

CloudPower97:
# Description

Part of #928

## Type of change

Instead of using `selectClassName` we now use `className`.
This has been done to make the various components homogeneous

# How Has This Been Tested?

The only test failing as of now is this one, but I'm not sure why:

```js
  test('calls onChange if provided', () => {
    const onChangeMock = jest.fn();
    wrapper = mount(
      <Select browserDefault onChange={onChangeMock} value="">
        <option value="" disabled>
          Choose your option
        </option>
        <option value="1">Option 1</option>
        <option value="2">Option 2</option>
        <option value="3">Option 3</option>
      </Select>
    );

    wrapper
      .find('option')
      .at(2)
      .simulate('click');

    expect(onChangeMock).toHaveBeenCalled();
  });
```

I've tried several times with different approaches, I'm at a dead end as of now.
But I can confirm the `component` correctly handles custom provided `onChange` function, as visible in `Storybook`, for example.

Any help is much appreciated!

# Checklist:


src/Select.js
```diff
@@ -1,141 +1,112 @@
...
+  error,
+  validate,
+  children,
+  multiple,
+  options,
+  value,
+  onChange,
+  ...props
+}) => {
+  const [selectedValue, setSelectedValue] = useState(value);
```
alextrastero: do we want to update the value if it's passed via props again?

test/Select.spec.js
```diff
@@ -39,21 +39,36 @@ describe('<Select />', () => {
...
+
+  //   wrapper
+  //     .find('option')
+  //     .at(2)
+  //     .simulate('click');
+
+  //   expect(onChangeMock).toHaveBeenCalled();
+  // });
+
+  // test('with icon', () => {
```
alextrastero: why is this test commented out?

CloudPower97: Whoops, that's my mistake

src/Select.js
CloudPower97: I think we should.
Otherwise, we might have a situation where the state is out of sync.
Think of `controlled component` pattern, the component will receive the `selected` value as `prop`.
What do you think?

CloudPower97: ping @alextrastero :smile:

test/Select.spec.js
```diff
@@ -39,16 +39,31 @@ describe('<Select />', () => {
...
+    wrapper = shallow(<Select validate className={'hello red'} />);
     expect(wrapper.find('select.hello.red.validate')).toHaveLength(1);
   });
 
-  test('calls onChange if provided', () => {
-    wrapper = shallow(<Select />);
-    wrapper.setState({ value: 'hello' });
-    expect(wrapper.find('select').prop('value')).toEqual('hello');
-  });
+  // test('calls onChange if provided', () => {
```
alextrastero: if you mount the component will this test work?

alextrastero:
Continued in #1074

PR #960, opened 2019-11-26
fix: remove waves effect when button is disabled

CloudPower97:
# Description

As per the [`materialize-css` example page](https://materializecss.com/pagination.html#!), disabled buttons do not have `waves`.

## Type of change

# How Has This Been Tested?

Currently, there is no `spec` for this component.
We might want to add at least one to check if setting `disabled` `prop` to true actually removes the `waves-effect` `css` class.

# Checklist:

### DIFF
--- a/src/PaginationButton.js
+++ b/src/PaginationButton.js
@@ -11,7 +11,7 @@ const PaginationButton = ({
   onSelect
 }) => {
   const classes = {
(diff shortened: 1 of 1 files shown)


CloudPower97:
ping @alextrastero :smile:

PR #974, opened 2019-12-02
The return of the info addon

alextrastero:
Restore addons


alextrastero:
The reason I had to try this was because a conflicting version of react.

PR #975, opened 2019-12-02
TextInput: convert to hooks:

felipebrunetti94:
- [x] refact to hooks;

See #928

### DIFF
--- a/src/TextInput.js
+++ b/src/TextInput.js
@@ -1,167 +1,154 @@
-import React, { Component } from 'react';
+import React, { useRef, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import idgen from './idgen';
 import constants from './constants';
 
-class TextInput extends Component {
-  constructor(props) {
-    super(props);
+const TextInput = ({
+  value,
+  children,
+  s,
+  m,
+  l,
+  xl,
+  disabled,
+  noLayout,
+  placeholder,
+  icon,
+  label,
+  inputClassName,
+  success,
+  error,
+  password,
+  email,
+  validate,
+  defaultValue,
+  type,
+  ...props
+}) => {
+  const getNnewId = id => {
+    id = id || idgen();
 
-    this.id = props.id || idgen();
-
-    if (props.password) {
-      this.id = `password${idgen()}`;
+    if (password) {
+      id = `password${idgen()}`;
     }
 
-    if (props.email) {
-      this.id = `email${idgen()}`;
+    if (email) {
+      id = `email${idgen()}`;
     }
+    return id;
+  };
+
+  const [newValue, setNewValue] = useState(value);
+
+  const inputRef = useRef(null);
+
+  useEffect(() => {
+    // eslint-disable-next-line react/prop-types
+    props['data-length'] && M.CharacterCounter.init(inputRef);
+  }, [props]);
+
+  useEffect(() => {
+    M.updateTextFields();
+  }, [value]);
+
+  const sizes = { s, m, l, xl };
+
+  let responsiveClasses;
+  if (!noLayout) {
+    responsiveClasses = { col: true };
+    constants.SIZES.forEach(size => {
+      responsiveClasses[size + sizes[size]] = sizes[size];
+    });
   }
 
-  componentDidMount() {
-    if (typeof M !== undefined) {
-      // eslint-disable-next-line react/prop-types
-      this.props['data-length'] && M.CharacterCounter.init(this.inputRef);
-    }
+  const wrapperClasses = cx('input-field', responsiveClasses);
+
+  let computedType;
+  if (type) {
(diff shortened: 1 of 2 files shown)


src/TextInput.js
```diff
@@ -1,167 +1,156 @@
...
-      computedType = 'password';
-    } else if (email) {
-      computedType = 'email';
-    } else {
-      computedType = 'text';
-    }
+  const inputProps = {
+    placeholder,
+    type: computedType,
+    id: newId,
```
talyssonoc: Shouldn't you be called the `newId` function here? Check the snapshot, it says the value of the id is a function and it should be a value, right?

src/TextInput.js
```diff
@@ -1,167 +1,156 @@
...
+  success,
+  error,
+  password,
+  email,
+  validate,
+  defaultValue,
+  type,
+  ...props
+}) => {
+  const newId = id => {
```
nolleto: Since this is a function, you could add a prefix to make more explicit that it is a function. Ex.:
- `getNewId`
- `createNewId`
- `generateNewId`
...

src/TextInput.js
```diff
@@ -1,167 +1,156 @@
...
+    return id;
+  };
 
-  componentDidMount() {
+  const [newValue, setNewValue] = useState(value);
+
+  const inputRef = useRef(null);
+
+  useEffect(() => {
     if (typeof M !== undefined) {
```
nolleto: This statement will always return `true` because `typeof` will always return a string.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof
> The **typeof** operator returns a string indicating the type of the unevaluated operand.

The right way would be `if (typeof M !== 'undefined') {`, but since this will change the behavior of this component would be better to check it with the repository owners.

test/__snapshots__/TextInput.spec.js.snap
```diff
@@ -12,7 +12,7 @@ exports[`<TextInput /> handles file type 1`] = `
     </span>
     <input
       className=""
-      id="mockId"
+      id={[Function]}
```
nolleto: 🚫

src/TextInput.js
CloudPower97: We should just drop this check altogether.
Let it throw an error if `M` is undefined

alextrastero: agreed, this check was wrong since the beginning since it just made the component fail silently

src/TextInput.js
alextrastero: yes, I would vote to move this to defaultProps, `id: idgen()`

alextrastero:
And thank you for the contribution! This component is one of the hard ones

src/TextInput.js
felipebrunetti94: @alextrastero , you say move the function to the idgen.js file or use the idgen() function ?
because if is the second option, the problem is to set again the ID in this situations:
`    if (password) {
      id = `password${idgen()}`;
    }

    if (email) {
      id = `email${idgen()}`;
    }`

felipebrunetti94: Sorry about the delay

felipebrunetti94:
@alextrastero is there anything else I can do to make it better ?

src/TextInput.js
```diff
@@ -1,167 +1,154 @@
...
+  success,
+  error,
+  password,
+  email,
+  validate,
+  defaultValue,
+  type,
+  ...props
+}) => {
+  const getNnewId = id => {
```
nolleto: Typo:  `getNewId`

alextrastero: how about instead of this logic we define `id` in default props?
```
id: `textinput-${idgen()}`
```

alextrastero: no real need for this password email logic

src/TextInput.js
```diff
@@ -1,167 +1,154 @@
...
+  };
+
+  const [newValue, setNewValue] = useState(value);
+
+  const inputRef = useRef(null);
+
+  useEffect(() => {
+    // eslint-disable-next-line react/prop-types
+    props['data-length'] && M.CharacterCounter.init(inputRef);
+  }, [props]);
```
alextrastero: the dependencies need to be more specific `props['data-length']`

src/TextInput.js
```diff
@@ -1,167 +1,154 @@
...
-    if (type) {
-      computedType = type;
-    } else if (password) {
-      computedType = 'password';
-    } else if (email) {
-      computedType = 'email';
-    } else {
-      computedType = 'text';
+    if (typeof icon === 'string') {
+      return <i className="material-icons prefix">{icon}</i>;
```
alextrastero: as per other components, we only accept `icon` as a node now, but you still need to append prefix: `return React.cloneElement(icon, { className: 'prefix' });`

Issue #977, opened 2019-12-04
RevealIcon throws warning in Card


aurinxki:
### Description
When rendering a Card, browser throws a warning of unrecognized property revealIcon

### Steps to Reproduce
`<Card
  className="my-class"
   title= "My title"
  actions={[]}
 >
  {children}
</Card>`

**Expected behavior:**
The component should render without warnings.

**Actual behavior:** 
Warning thrown in browser's console when rendering:
React does not recognize the `revealIcon` prop on a DOM element.

### Versions

**react-materialize:**3.4.1 
**materialize-css:** 1.0.0
**react:** 16.10.2

abnersajr:
PR opened fixing this: https://github.com/react-materialize/react-materialize/pull/978
`revealIcon`has a default value and this is being injected in a `div` causing this warnings.

PR #978, opened 2019-12-05
Fix Card with revealIcon

abnersajr:
# Description

This PR fixes the issue #977 where some issues are beign caused because of the wrong injection of some props.

The `Card` component has two props that was beign injected in a `div` and it's causing the console warning.
`closeIcon` and `revealIcon` isn't valid attributes to a `div` tag. Those props are being injected using the spread technique. 
Those props aren't used in the render scenario, and this was the motivation to use the `delete` way to handle it.

Also was fixed an issue in the code where the `revealIcon` was evaluated as an object. The component was not working correctly and was fixed.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Open the storybook you see the first issue with the invalid `object` in the DOM.
Having storybook opened and chrome devtools console you can see the warning about the invalid props.
Digging into the DOM tree you see the two invalid attributes `revealicon & closeicon` what makes the errors appears.

# Checklist:

### DIFF
--- a/test/__snapshots__/Card.spec.js.snap
+++ b/test/__snapshots__/Card.spec.js.snap
@@ -3,16 +3,6 @@
 exports[`<Card /> renders 1`] = `
 <div
   className="blue-grey darken-1 card"
-  closeIcon={
-    <Icon>
-      close
-    </Icon>
-  }
-  revealIcon={
-    <Icon>
-      more_vert
-    </Icon>
-  }
 >
   <div
     className="card-content white-text"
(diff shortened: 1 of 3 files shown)


alextrastero:
Fixes #977

test/__snapshots__/Card.spec.js.snap
```diff
@@ -3,16 +3,6 @@
...
-  closeIcon={
-    <Icon>
-      close
-    </Icon>
-  }
-  revealIcon={
-    <Icon>
-      more_vert
-    </Icon>
-  }
```
alextrastero: how come this was removed from the snapshot?

alextrastero:

PR #979, opened 2019-12-13
Navbar centerChildren

ArtiomLK:
# Description
The Navbar children will be constrained in a container rather than spread out to the far edges.
 
I wanted to center the NavBar items as I previously did when using the materializecss library
navbar

By just adding the centerChildren property it allows us to bound the Navbar items in a container, thus the Navbar items will be aligned in a container rather than spread out to the far edges

#### View:
docs Navbar view

#### Info:
docs Navbar info

## Type of change

# How Has This Been Tested?

#### Navbar.spec.js
jest Navbar

#### All tests pass:
npm run test

# Checklist:

### DIFF
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -28,6 +28,7 @@ class Navbar extends Component {
       fixed,
       alignLinks,
       centerLogo,
+      centerChildren,
       search,
       menuIcon,
       sidenav
@@ -61,7 +62,9 @@ class Navbar extends Component {
 
     let navbar = (
       <nav className={navCSS}>
-        <div className="nav-wrapper">
+        <div
+          className={centerChildren ? 'nav-wrapper container' : 'nav-wrapper'}
+        >
           {search ? (
             <SearchForm />
           ) : (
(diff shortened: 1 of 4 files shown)


alextrastero:
Very well structured PR, thanks

src/Navbar.js
```diff
@@ -61,7 +62,9 @@ class Navbar extends Component {
 
     let navbar = (
       <nav className={navCSS}>
-        <div className="nav-wrapper">
+        <div
+          className={centerChildren ? 'nav-wrapper container' : 'nav-wrapper'}
```
alextrastero: question, instead of adding a new interface prop, can't we add `container` as a className?
`<Navbar className='container' />`

ArtiomLK: The benefit of adding an interface prop is that the users would have the choice to centered or not their Navbar child components as shown in the conversation images. If we set the class name to container, `<div className="container">` it will only show the Navbar child components centered.

I tested it by applying the recommendation with `alignLinks="left"` and the #981 fix, which 
rendered as:
actual



 rather than:
expected

ArtiomLK:
I would strongly suggest checking the #981 PR as well since that PR contains a critical bug fix that complements this approved PR

PR #994, opened 2020-02-21
Fix undefined check

marr:
# Description

Dropdown was throwing because it was trying to invoke a method on `M` when `M` was undefined. Components don't work as expected without the `M` global defined. This should be added to the documentation:

`window.M = require('materialize-css')`

Without that, things like dropdowns don't work.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)


alextrastero:
Thanks for the contribution!

PR #995, opened 2020-02-21
TextInput: converted to use hook

DangerN:
# Description

Part of #928 

# How Has This Been Tested?

Passed tests

# Checklist:

### DIFF
--- a/src/TextInput.js
+++ b/src/TextInput.js
@@ -7,8 +7,6 @@ import constants from './constants';
 const TextInput = props => {
   const [id, setId] = useState(props.id || idgen());
 
-  const [idSetable, setIdSetable] = useState(true);
-
   const inputRef = useRef(null);
 
   // eslint-disable-next-line react/prop-types
--- a/src/TextInput.js
+++ b/src/TextInput.js
@@ -9,25 +9,13 @@ const TextInput = props => {
 
   const [idSetable, setIdSetable] = useState(true);
 
-  if (props.password && idSetable) {
-    setId(`password${idgen()}`);
-    setIdSetable(false);
-  }
-
-  if (props.email && idSetable) {
-    setId(`email${idgen()}`);
-    setIdSetable(false);
-  }
-
   const inputRef = useRef(null);
 
   // eslint-disable-next-line react/prop-types
   const dataLength = props['data-length'];
 
   useEffect(() => {
-    if (typeof M !== undefined) {
-      dataLength && M.CharacterCounter.init(inputRef);
-    }
+    dataLength && M.CharacterCounter.init(inputRef);
   }, [dataLength]);
 
   useEffect(() => {
@@ -239,11 +227,7 @@ TextInput.propTypes = {
   /*
    * email type
    */
-  email: PropTypes.bool,
-  /*
-   * dataLength type
-   */
-  dataLength: PropTypes.number
+  email: PropTypes.bool
 };
 
 export default TextInput;
--- a/src/TextInput.js
+++ b/src/TextInput.js
@@ -1,11 +1,11 @@
-import React, { useRef, useState, useEffect } from 'react';
+import React, { useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import idgen from './idgen';
 import constants from './constants';
 
 const TextInput = props => {
-  const [id, setId] = useState(props.id || idgen());
+  const id = props.id || idgen();
 
   const inputRef = useRef(null);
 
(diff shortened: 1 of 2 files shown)


src/TextInput.js
```diff
@@ -1,167 +1,161 @@
...
+  }, [dataLength]);
 
-  componentDidUpdate(prevProps) {
-    const { value } = this.props;
-
-    if (value !== prevProps.value && typeof M !== 'undefined') {
+  useEffect(() => {
+    if (typeof M !== 'undefined') {
       M.updateTextFields();
     }
```
alextrastero: we've been removing this check on other components, since with this check it will silently fail. Best if just `M.updateTextFields()`

src/TextInput.js
```diff
@@ -245,7 +239,11 @@ TextInput.propTypes = {
   /*
    * email type
    */
-  email: PropTypes.bool
+  email: PropTypes.bool,
+  /*
+   * dataLength type
+   */
+  dataLength: PropTypes.number
```
alextrastero: let's remove this prop, we are only using `data-length` as a data attribute, but no need to document it here

DangerN:
Thanks for the feedback, I've made the changes you requested.

src/TextInput.js
```diff
@@ -1,167 +1,149 @@
...
+  useEffect(() => {
+    dataLength && M.CharacterCounter.init(inputRef);
+  }, [dataLength]);
 
-    if (props.email) {
-      this.id = `email${idgen()}`;
+  useEffect(() => {
+    if (typeof M !== 'undefined') {
+      M.updateTextFields();
     }
```
alextrastero: as mentioned, let's remove this check as we want it to fail with reason if M is not defined :)

src/TextInput.js
```diff
@@ -1,167 +1,149 @@
...
 import constants from './constants';
 
-class TextInput extends Component {
-  constructor(props) {
-    super(props);
+const TextInput = props => {
+  const [id, setId] = useState(props.id || idgen());
 
-    this.id = props.id || idgen();
+  const [idSetable, setIdSetable] = useState(true);
```
alextrastero: not needed anymore?

src/TextInput.js
DangerN: I could have sworn I got that one :P

DangerN: Also I have removed the test that asserts it wont fail if M is not defined.

src/TextInput.js
DangerN: We are still using `props.id || idgen()` to set the id on line 8. Line 10 is used to ensure that the id can only be set once per mount.

DangerN:
Ok I think we should be good now. Also would you mind explaining this 'global M' to me? I looked through the code but couldn't figure out where it was coming from.

src/TextInput.js
alextrastero: I'm not sure about this, it seems that neither `idSetable` nor `setIdSetable` are used anywhere right?

DangerN: You're absolutely right, it was removed in a previous edit. Also after thinking about it some more I understand why the associated logic was unnecessary.

DangerN:
Thank you for taking the time to review my code and answer my questions. So after thinking some more about how `M` got there, does this mean that when developing with node that all dev dependancies from packages are included in the dev environment using said package?  

And that a production build using webpack or something similar will build from them?

If thats the case why even specify a difference between dependency types?

If you could suggest some reading on the topic I would really appreciate it.

alextrastero:
In this particular case it doesn't relate to dependencies.
Here we are importing a script file (`<script src="/path-to-materialize.js" />`), this file once imported gets executed and it must have something around the lines of:

```js
var M = {
  Modal: () => {},
  Overlay: () => {},
  ...
}

window.M = M;
```

That's why you can access `M` object from anywhere in the App, because it's added to window.

alextrastero:
A piece of advice when writing commit messages, these messages in most cases are considered reasons as to why changes are made, i.e. `upgrade react dependency` or `fix Modal is undefined error`.
One day down the road, someone will look at a line of code and say "why was this done?", then running `git blame` it will be visible that the commit said `fixed`.

Do you see what I mean? :)

Thanks so much for the contribution @DangerN, please keep them coming!

DangerN:
Thanks for the clarification. And I will make my commit messages more descriptive, I do know better 🙃

src/TextInput.js
```diff
@@ -1,167 +1,145 @@
...
-
-    if (props.password) {
-      this.id = `password${idgen()}`;
-    }
-
-    if (props.email) {
-      this.id = `email${idgen()}`;
-    }
+const TextInput = props => {
+  const id = props.id || idgen();
```
alextrastero: I'm going to merge this since the task in hand is 99% done, although here we have a potential bug, every time this component rerenders it will create another `id` and this is not desired, if you want to overwrite the id prop it should live in `defaultProps` I will create a follow up PR and link it here

PR #996, opened 2020-02-24
Support false/null elements in Dropdown

marr:
# Description

Dropdown was failing to support null/false elements due to an incompatible type check. The check was not checking if an element existed before inspecting `element.type.name`. Standard elements have a type of "a", so the check was passing, but null elements don't have a type therefore need to have a truthy check before inspecting.

## Type of change


alextrastero:
Nice! thx

PR #1086, opened 2020-04-01
Modal document fix

### LINKED ISSUE #1085: Modal 3.7.x break in Next.js
### Description

Since Next.js is SSR, it will not recognize the document

Need to modify this line 34 on Modal.js

```var MODAL_CONTAINER = document.body;```

into 

```var MODAL_CONTAINER = (typeof window !== 'undefined')  ? document.body : null;```

I know, the modal might not work in Next.js anyway, but I still can use the rest of components in react-materialize without breaking.

### PR DESCRIPTION
**Pull request by alextrastero**
Fixes #1085

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,6 +19,7 @@ globals:
   global: true
   window: true
   module: true
+  Element: true
 
 plugins: ['prettier', 'react-hooks']
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 3.8.0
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by joecapra** [comment]
New at react so please forgive me. How do i implement this. the example show root={[object HTMLBodyElement]}. using that gives an error. im using create-react-app and my root is 'root' so what do i need to change root={[object HTMLBodyElement]} to in order to get it to work?

**Comment by alextrastero** [comment]
```
  /**
   * Root node where modal should be injected
   * @default document.body
   */
```

So for your case it should `document.getElementById('root')`

PR #1105, opened 2020-04-20
(re)added onChange and defaultValue to Textarea

tomatenschaf:
# Description

onChange and defaultValue props were missing in rendered <textarea /> since last refactoring of Textarea (about 22 day ago). 

## Type of change

# How Has This Been Tested?

Added tests to avoid this bug in future.

# Checklist:


alextrastero:
Looking good! could you remove the changelog and lib folder if possible?

alextrastero:
I will handle that when building a new version

alextrastero:
ping

alextrastero:
ping

tomatenschaf:
I reverted my libfolder and changelog commits. Sorry, I didn't realized, an action was required.

src/Textarea.js
```diff
@@ -55,7 +55,9 @@ const Textarea = ({
         placeholder={placeholder}
         id={id}
         value={value}
+        defaultValue={defaultValue}
         ref={textareaRef}
+        onChange={onChange}
```
alextrastero, line 60: nice!, since we're using `...rest` how about delete L21 and L23 and that's it?

tomatenschaf: I prefer the explicit list of parameters. `...rest` is for HTMLTextareaElement attributes, that have no special meaning in react or materialize.

alextrastero: fair enough

alextrastero:
Fixes #1115

Issue #880, opened 2019-06-07
Calling createRoot in constructor of Modal causes SSR applications to fail


tm1000:
### Description

Issue is with the calling of createRoot in Modal.jsx. It happens in the constructor on line 13 https://github.com/react-materialize/react-materialize/blob/master/src/Modal.js#L13

For SSR (server side rendering) applications createRoot can't be called because document does not exist. Thus it must be called inside of componentDidMount instead. Looking over this visually it doesn't appear to be an issue if it's in there

### Steps to Reproduce

1. Create a SSR (ca use next.js or something else)
2. add a modal
3. page fails on server render with "document is undefined" (because it is)

**Expected behavior:** Modal should technically render on the server but could also render in componentDidMount

**Actual behavior:** "document is undefined" error

### Versions

**react-materialize:** 3.3.1
**materialize-css:** 1.0.0
**react:** 16.8.6

PR #1106, opened 2020-04-22
Fix broken start of Modal on Next JS

sunaryohadi:
# Description

This will avoid breaking while loading on SSR (NextJS). 

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [ x ] My changes generate no new warnings
- [ x  ] I have not generated a new package version. (the maintainers will handle that)

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -194,7 +194,7 @@ Modal.propTypes = {
 
 Modal.defaultProps = {
   id: `Modal-${idgen()}`,
-  root: document.body,
+  root: typeof window !== 'undefined' ? document.body : null,
   open: false,
   options: {
     opacity: 0.5,


randytarampi:
Just in case someone's looking for a packaged version, I've published this for my own use in the meantime: https://www.npmjs.com/package/@randy.tarampi/react-materialize.

Fixes a the original issue, https://github.com/react-materialize/react-materialize/issues/1085, and something a bit older too, https://github.com/react-materialize/react-materialize/issues/880.

alextrastero:
Thanks for the proposal, but if no body is null the output is:
`Cannot read property 'appendChild' of null`

alextrastero:
This should be fixed via in `3.8.6`, thanks again for reporting this and submitting a fix!

GnarusGBaby:
I am using version 3.9.6 and still having the issue.

Issue #989, opened 2020-01-13
Radio input id's are duplicated


aurinxki:
### Description
If more than one radio groups are found in the same DOM, the same ID's are assigned to its options, preventing the correct handling of events.

### Steps to Reproduce
1. Render two RadioGroup
```
<Fragment>
	<RadioGroup
		onChange={this.handleFirstChange}
		options={[
			{ label: 'No', value: 0 },
			{ label: 'Yes', value: 1 }
		]}
	/>
	<RadioGroup
		onChange={this.handleSecondChange}
		options={[
			{ label: 'Other No', value: 3 },
			{ label: 'Other Yes', value: 4 }
		]}
	/>
</Fragment>
```

2. Select an option for first radio group
3. Select an option for second radio group

**Expected behavior:** 
The onChange function given to the Select is triggered and the component's value is updated.

**Actual behavior:** 
For the first radiogroup, it's options get the following id's: #radio0 and #radio1. 
For the second , it's options are assigned the same id's: #radio0 and #radio1. 

Because of DOM precedence, the first incidence of #radio0 is found and its event (handleFirstChange) is triggered, instead of (handleSecondChange).

**More information:**
react-materialize radio group doesn't receive a parameter for custom id. 
When creating the component, it maps its options and uses the index as the id. 
Every time the constructor builds a new instance, it starts on 0.
```
return _react.default.createElement(_react.default.Fragment, null, options.map(function (radioItem, idx) {
    return _react.default.createElement("label", {
      htmlFor: "radio".concat(radioItem.name).concat(idx),
      key: "radio".concat(idx)
    }, _react.default.createElement("input", {
      id: "radio".concat(radioItem.name).concat(idx),
    }
```

### Versions

**react-materialize:** 3.4.1
**materialize-css:** 1.0.0
**react:** 16.10.2

hb0:
Same here on:

# Versions

**react-materialize:** 3.9.2 (and 3.5.9)
**materialize-css:** 1.0.0
**react:** 16.13.1

wcomicho:
facing the same problem here.
When I click item2, item1 gets updated and not item2
`
 ...
const[formValues, setFormValues] = useState({item1: 'yes', item2: 'yes'});
...
 //On change event of all Radio Group
 const handleChange = (e) =>{
    setFormValues({...formValues, [e.target.name]: e.target.value});
  }
 ...
<RadioGroup
                id="item1"
                name="item1"
[...]

PR #1134, opened 2020-06-23
Add 'id' parameter to RadioGroup to fix issue #989

hb0:
# Description

This is a very simply approach to fix the issue #989 .
* Currently the `id`s of each radio `<input>` are named as follows: `radio0`, `radio1`, ...
* Because of this the `id`s of different `<RadioGroup>`s are non-unique
* Thus, the `onChange` handler of the first `<RadioGroup>`  is always called, also when a second group was added and clicked

With this PR the `radio` `id`s get unique names, based on a provided `id`-parameter:
* without `id` parameter in the `<RadioGroup>` the `id`s names are still generated as before (backward-compatibility)
* with `id` parameter in the `<RadioGroup>` the `id`s between multiple groups are unique: `RadioGroup_${id}_radio0`

## Type of change

# How Has This Been Tested?

1. All tests run through: `npm test a`.
2. Backward-compatibility is ensured: The existing test case `onChange callback` in `RadioGroup.spec.js` was adjusted to fail if the ids are not generated as before.
3. Added a second test case `onChange callback with id parameter` which makes sure a provided `id` parameter is used to generate a unique ids.
4. I added the altered version of the `RadioGroup` component to my own project were I had the issue #989.
With the new `RadioGroup` component and after adding the `id` parameter to the RadioGroups the issue #989 is fixed:
```
<RadioGroup
          id="groupA"
          withGap
          name="groupA"
          label="Group A"
          options={[
            {
              label: 'Option 1',
              value: 'option1'
            },
            {
              label: 'Option 2',
              value: 'option2'
            }
          ]}
          onChange={this.testHandlerA}
        /><br /><br /><br />

        <RadioGroup
          id="groupB"
          withGap
          name="groupB"
          label="Group B"
          options={[
            {
              label: 'Option 1b',
              value: '1b'
            },
            {
              label: 'Option 2b',
              value: '2b'
            }
          ]}
          onChange={this.testHandlerB}
        />
```

i.e. clicking on ` 'Option 2b',` now triggers `testHandlerB` with the correct `event.target`.
Before the `testHandlerA` was called with `option2` as event.target.

# Checklist:

### DIFF
--- a/src/RadioGroup.js
+++ b/src/RadioGroup.js
@@ -9,18 +9,21 @@ const RadioGroup = ({
   name,
   radioClassNames,
   value,
-  options
+  options,
+  id
 }) => {
+  // Generates unique radio-ids if an 'id' is provided to fix issue #989
+  const idPrefex = typeof id === 'undefined' ? `` : `RadioGroup_${id}_`;
   return (
     <React.Fragment>
       {options.map((radioItem, idx) => (
         <label
           className={radioClassNames}
-          htmlFor={`radio${idx}`}
-          key={`radio${idx}`}
+          htmlFor={`${idPrefex}radio${idx}`}
+          key={`${idPrefex}radio${idx}`}
         >
           <input
-            id={`radio${idx}`}
+            id={`${idPrefex}radio${idx}`}
             value={radioItem.value}
             type="radio"
             checked={radioItem.value === value}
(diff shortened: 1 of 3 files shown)


alextrastero (commented):
Thx for this, but how about we use the same approach as with for example `Textarea` L166?

hb0:
Should work - and so much simpler :+1: 

When I created the PR I was very new to React and was just happy to find a temporary solution to code on ;-)

PR #1144, opened 2020-08-10
Preloader fix size class name

WaeCo:
Preloader component applied invalid size class name.
It was always "*size*" instead of the value of the `size` property.
This fixes this issue and uses the correct value.


alextrastero (approved):
Thanks!

PR #1145, opened 2020-08-10
Fix Spinner not exported

WaeCo:
This exports the Spinner component as it is already exported in the typescript definitions.


alextrastero:
Super thanks!

Issue #922, opened 2019-11-04
Icon and font-size


lynicidn:
### Description
> 
> Docs: https://materializecss.com/icons.html
> To control the size of the icon, change the font-size: 30px

### Steps to Reproduce
try it:
```html
<Icon style={{fontSize: "30px"}}>lock</Icon>
```
**Expected behavior:** icon with 30px of font size

**Actual behavior:** font size 24px from `material-icon` class of css

### Versions

materialize-css@ (empty)
react-materialize@3.4.1 
react@16.11.0

lynicidn:
i think problem here:
https://github.com/react-materialize/react-materialize/blob/master/src/Icon.js#L18

we should transfer `style` for `i` component

alextrastero:
@lynicidn good catch, we should maybe be passing `...rest` props:
`<i className={cx(classes, this.props.className)} {...rest} />`

Is this something you would like to give a go?

lynicidn:
@alextrastero sorry, i can't.
1. bad know english
2. beginer in react/jsx 
3. don't understand about `rest` - it transformed props? as example: `fontSize` to `font-size` ?

i find temporary solve via className - copy css class `large` to `giant` and set Font Size for this class

aelfric:
Is there still interest in fixing this?  I could submit a PR to add the spread props like you suggested.

alextrastero:
Yes please!

PR #1153, opened 2020-10-05
fix: pass HTML element props to `i` tag for Icon

aelfric:
# Description

Addresses #922 by passing HTML Element props to the `i` tag created by `Icon`.

## Type of change

# How Has This Been Tested?

Added a test case to `Icon.spec.js` and also independently updated the `Icon.d.ts` and verified that the following is valid:

```typescript
    let props: IconProps  ={
        className: "test",
        left: true,
        large: true,
        style: {
            fontSize: "30px"
        },
        onClick: ()=>undefined,
    }
```

But something like this is still invalid:
```typescript
    let props: IconProps  ={
        className: "test",
        left: true,
        large: true,
        style: {
            fontSize: "30px"
        },
        onClick: ()=>undefined,
        // Prop that isn't permitted
        foo: ""
    }
```
# Checklist:

### DIFF
--- a/src/Icon.js
+++ b/src/Icon.js
@@ -3,17 +3,24 @@ import PropTypes from 'prop-types';
 import constants from './constants';
 import cx from 'classnames';
 
-const Icon = props => {
+const Icon = ({ className, children, ...props }) => {
   const classes = {
     'material-icons': true
   };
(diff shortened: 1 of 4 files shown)


alextrastero (approved):
Great! Thank you so much, keep em coming!

PR #1155, opened 2020-10-09
fix: need to add 'invalid' class for errors to be displayed

### LINKED ISSUE #1132: Text Area Custom Error String is not rendered in the DOM
### Description
Text Area Custom Error String is not rendered in the DOM
### Steps to Reproduce
Send Prop error as add a custom string to display an error message
 It updated as a span Tag with error but does not render on screen

aelfric:
# Description

Materialize will only display the error text on a form input if the `invalid` class is present on the underlying `input` element.  
Fixes #1132 for both`Textarea` and `TextInput`.  I'm not sure if there are other components where this fix should be applied.

## Type of change

# How Has This Been Tested?

Storybook stories were added to check the visual display of the error messages and new test cases added to the Textarea and TestInput specs to verify that the class is being added.

# Checklist:

### DIFF
--- a/src/TextInput.js
+++ b/src/TextInput.js
@@ -129,7 +129,7 @@ const TextInput = React.forwardRef((props, ref) => {
       {renderIcon()}
       <input
         ref={inputRef}
-        className={cx({ validate }, inputClassName)}
+        className={cx({ validate }, { invalid: error }, inputClassName)}
         {...inputProps}
       />
       {renderLabel()}
--- a/src/Textarea.js
+++ b/src/Textarea.js
@@ -59,7 +59,12 @@ const Textarea = ({
         ref={textareaRef}
         onChange={onChange}
         {...rest}
-        className={cx('materialize-textarea', { validate }, className)}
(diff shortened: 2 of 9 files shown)


alextrastero:
I'm not sure this is actually needed, materialize-css will add this class for you, check out how to display error messages here:
https://codesandbox.io/s/elegant-hooks-8hr5w?file=/src/App.js

aelfric:
In your example, the `invalid` class is automatically added because the input has type `email`, which has validation built into it.  I believe the creator of the original issue was trying to put error messages on general text fields.

alextrastero:
> My change would break the existing functionality on type='email'.

it would or wouldn't?

aelfric:
This fix is wrong.  It works for plain text fields but does not work for ones that actually do have validation logic, like email fields.  Do you agree that adding a prop to manually indicate something is invalid makes sense?  If so, I'll update my PR.

aelfric:
The original fix was wrong and I never had the chance to revisit this.

Issue #1167, opened 2020-12-14
Dropdown with inline JS issue in 3.9.4


Tchekda:
### Description

After upgrading to v3.9.4, I faced a major problem with my dropdown.

### Steps to Reproduce

1. Update to v3.9.4
2. Create a Dropdown looking like this one : 
(You can obviously use any bool condition instead of `hasRole`)
```js
<Dropdown
  id="security-dropdown"
  trigger={
    <a href="#!">
      Sécurité <Icon right>arrow_drop_down</Icon>
    </a>
  }>
  {hasRole('ROLE_GUARDIAN') ? (
    <>
      <Link to={`/${userType}/email`}>Adresse Mail</Link>
      <Link to={`/${userType}/other`}>Autre</Link>
    </>
  ) : (
    <Link to={`/${userType}/password`}>Mot de Passe</Link>
  )}
</Dropdown>
```

3. Admire this beautiful error : `TypeError: (intermediate value).filter is not a function` located at `renderItems
xxxxx/node_modules/react-materialize/lib/Dropdown.js:54`

![image](https://user-images.githubusercontent.com/23559888/102145395-6a3f0280-3e67-11eb-99df-b12657445f5d.png)

**Expected behavior:** Display correctly the dropdown as it was working in v3.9.3

**Actual behavior:** The temporary fix that I found is to make to separate conditions. 
```js
<Dropdown
  id="security-dropdown"
  trigger={
    <a href="#!">
      Sécurité <Icon right>arrow_drop_down</Icon>
    </a>
  }>
  {hasRole('ROLE_GUARDIAN') && (
    <Link to={`/${userType}/email`}>Adresse Mail</Link>
  )}
  {hasRole('ROLE_GUARDIAN') && (
    <Link to={`/${userType}/other`}>Autre</Link>
  )}
  {!hasRole('ROLE_GUARDIAN') && (
    <Link to={`/${userType}/password`}>Mot de Passe</Link>
  )}
</Dropdown>
```

**Explanations :** The PR #1164 made by @MauAbata changed the way Dropdowns are rendered, that had impact on my code, and I suppose and many others.

### Versions

**react-materialize:** 3.9.4
**materialize-css:** 1.0.0
**react:** 17.0.1

Tchekda:
And the is a second error that is linked to the first : 
`TypeError: can't access property "type", element is null` located at `renderItems/<
[...]

PR #1164, opened 2020-12-08
Allow Dropdown to accept null, li, and Divider children.

MauAbata:
# Description

This change allows the `<Dropdown/>` component to accept `null` and `li` items, as well as fixes rendering of dropdowns in minified builds (per #993).

Accepting `null` items allows for conditional item semantics, such as:

```jsx
const isActive = false;
<Dropdown>
  { isActive && <a href={'...'}>Deactivate</a> }
  <li>
    <a>Do not wrap this li</a>
  </li>
  <Divider />{/* <-- now renders without an LI in minified code */}
</Dropdown>
```

Prior to this change, `null` children are not excluded from processing, and raise errors inside the Dropdown component.

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Minimal viable example is provided above, code change is trivial and has been tested in minified code builds.

# Checklist:

- [?] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have not generated a new package version. (the maintainers will handle that)

PR #1168, opened 2020-12-15
Allow for Dropdown children to also be Fragments.

MauAbata:
# Description

This is a followup to #1164 which addresses the issue with Fragments in dropdowns, as discussed in #1167. Now, the code will recurse on fragments, rendering each properly into their own `<li>`. This retains the functionality of allowing null and undefined components.

The previous PR broke on Fragments because in the example in the issue, it was presented as a single-item child and the code assumed `children` would only ever be an array. **That PR will also break on other single-item dropdowns!**

## Type of change

# How Has This Been Tested?

Test with components similar to these:

```jsx
// Multi-element with boolean values, Divider, and li test:
<Dropdown>
  { false && <a>Do Not Show Me</a> }
  <a>Show Me</a>
  <Divider/>
  <li><a>No Additional Wrapper</a></li>
</Dropdown>

// Single element render:
<Dropdown>
  <Divider/>
</Dropdown>

// Fragment, multi-element:
<Dropdown>
  <React.Fragment>
    <a>One</a>
    <a>Two</a>
  </React.Fragment>
  <a>Three</a>
</Dropdown>

// Fragment, only child:
<Dropdown>
  <React.Fragment>
    <a>Fragment</a>
  </React.Fragment>
</Dropdown>
```

# Checklist:

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -29,16 +29,22 @@ const Dropdown = ({ children, className, trigger, options, ...props }) => {
       className: cx(trigger.props.className, 'dropdown-trigger')
     });
 
-  const renderItems = () =>
-    Children.map((children || []).filter(Boolean), element => {
-      if (element.type === 'li') {
+  const renderItems = c =>
+    Children.map(c, element => {
+      if (!element || element.type === 'li') {
+        // Skip null, undefined, or li children
         return element;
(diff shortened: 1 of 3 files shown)


alextrastero (approved):
Yup, looks good, thanks!

PR #1171, opened 2021-01-06
Make onSelect on CollapsibleItem optional (TypeScript)

7coil:
This change partially reverts the changes made in 4308d86fbc99e28eaff7d676c44a18d1c823be6d, which made `onSelect` on `CollapsibleItem` mandatory.

# Description

`onSelect` should not be a mandatory prop of `CollapsibleItem`. [Existing documentation do not use the prop](http://react-materialize.github.io/react-materialize/?path=/story/javascript-collapsible--popout), and not including it flags a type error.

image

## Type of change

# How Has This Been Tested?

Cloned and built whilst checking that the build process emits `onSelect?: (eventKey: any) => any;` into the `CollapsibleItem.d.ts` file.
Also checked the typing error no longer occurs.

image

# Checklist:

This change (probably) does not warrant any commenting.

### DIFF
--- a/types/components/CollapsibleItem.d.ts
+++ b/types/components/CollapsibleItem.d.ts
@@ -7,7 +7,7 @@ export interface CollapsibleItemProps extends SharedBasic {
   header: any;
   icon?: React.ReactNode;
   iconClassName?: string;
-  onSelect: (eventKey: any) => any;
+  onSelect?: (eventKey: any) => any;
   eventKey?: any;
 }
 


7coil:
(N.B. Building on Windows required a change in the "copy-types" script in package.json -- Might be helpful if this was multiplatform by doing the copy process inside a js script)

alextrastero:
Thanks @7coil! solved in #1172

